PR #6, opened 2024-04-02
Update README.md - Fix broken links

farouqaldori:
There are some broken links in the readme file, I fixed them to point to the right folders.


klieret:
@carlosejimenez @john-b-yang Gonna merge this since it's a fix and shouldn't interfere with anything

PR #12, opened 2024-04-02
chore: remove gnureadline dependency

MontagueM:
requirement is not used anywhere and causes a setup failure on windows as the dependency cannot be used with windows


klieret (approved):
LGTM! Thanks!

Issue #14, opened 2024-04-02
Azure ecosystem support


cpendery:
It would be great if the agent supported the Azure ecosystem both through supporting Azure OpenAI & through pulling Azure Devops stories/tasks instead of Github issues.

ofirpress:
If you submit a PR we could consider it but as this is a repo for quick & efficient research I don't think that this will be on our priority list for now.

cpendery:
@ofirpress that's fair, I'm happy to submit a set of PRs for it

PR #16, opened 2024-04-03
feat: add support for azure openai

cpendery:
Adds support for models hosted on Azure OpenAI under the model prefix `azure:`. 

Part of #14 & #22

### DIFF
--- a/sweagent/agent/models.py
+++ b/sweagent/agent/models.py
@@ -224,8 +224,8 @@ class OpenAIModel(BaseModel):
         # Set OpenAI key
         cfg = config.Config(os.path.join(os.getcwd(), "keys.cfg"))
         if self.args.model_name.startswith("azure"):
-             self.api_model = cfg["AZURE_OPENAI_DEPLOYMENT"]
-             self.client = AzureOpenAI(api_key=cfg["AZURE_OPENAI_API_KEY"], azure_endpoint=cfg["AZURE_OPENAI_ENDPOINT"], api_version=cfg.get("AZURE_OPENAI_API_VERSION", "2024-02-01"))
+            self.api_model = cfg["AZURE_OPENAI_DEPLOYMENT"]
+            self.client = AzureOpenAI(api_key=cfg["AZURE_OPENAI_API_KEY"], azure_endpoint=cfg["AZURE_OPENAI_ENDPOINT"], api_version=cfg.get("AZURE_OPENAI_API_VERSION", "2024-02-01"))
         else:
             self.client = OpenAI(api_key=cfg["OPENAI_API_KEY"])
 
--- a/sweagent/agent/models.py
+++ b/sweagent/agent/models.py
@@ -225,8 +225,7 @@ class OpenAIModel(BaseModel):
         cfg = config.Config(os.path.join(os.getcwd(), "keys.cfg"))
         if self.args.model_name.startswith("azure"):
              self.api_model = cfg["AZURE_OPENAI_DEPLOYMENT"]
-             api_version = cfg.get("OPENAI_API_VERSION", "2024-02-01")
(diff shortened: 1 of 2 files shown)


cpendery:
> Thanks for opening this PR.
> 
> This seems to introduce a lot of code duplication. Couldn't you just subclass `OpenAIModel`? Or perhaps even better to add some logic in `OpenAIModel.__init__` to just initialize `self.client` according to the prefix of `SHORTCUTS`?

I’m happy to do either, I was just trying to avoid changing the convention of one class extending the base per model provider. I’ll go with the updating the OpenAIModel

klieret (approved):
LGTM! If you confirm that you've tested it, I think we're good to merge :)

cpendery:
> LGTM! If you confirm that you've tested it, I think we're good to merge :)

If someone else can validate it before we merge, that would be great. I'm only on Windows so all the other issues are blocking me for verifying this works for the whole pipeline

klieret:
Let me ping @0xdevalias and @kim-borgen since they upvoted. Perhaps they can test this.

0xdevalias:
> since they upvoted. Perhaps they can test this.

@klieret I'm not currently setup to be able to test this unfortunately. I just liked the idea of wider support.

kim-borgen:
@klieret Unfortunately I only have the 8k context length available for testing now and the first prompt (messages) is 10,963. Other than that, it seems to work. Sorry but I can't dedicate more time, but if someone supplies a config that uses less than 6k tokens for example I can test quickly 

Otherwise: 
- Requested change: add api_version to cfg keys
- When trying to get the API working I get this error, so I had to delete the traj/kim folder every time, a bit annoying. 
```
WARNING  **************************************************                                                                                                                                                                                                                                                                                                                                                              
WARNING  Found existing args.yaml with different arguments!                                                                                                                                                                                                                                                                                                                                                              
WARNING  **************************************************   
```

klieret:
@kim-borgen Thanks for the quick check!  Can you open issues with some more context for the "Otherwise" ;)  (not sure if I fully understand the context of the second one yet)

sweagent/agent/models.py
```diff
@@ -220,7 +223,11 @@ def __init__(self, args: ModelArguments, commands: list[Command]):
 
         # Set OpenAI key
         cfg = config.Config(os.path.join(os.getcwd(), "keys.cfg"))
-        self.client = OpenAI(api_key=cfg["OPENAI_API_KEY"])
+        if self.args.model_name.startswith("azure"):
+             self.api_model=cfg["AZURE_OPENAI_DEPLOYMENT"]
+             self.client = AzureOpenAI(api_key=cfg["AZURE_OPENAI_API_KEY"], azure_endpoint=cfg["AZURE_OPENAI_ENDPOINT"], api_version="2024-02-01")
```
mspronesti, line 228: How about parametrizing the `api_version` too? It changes rather frequently :)
```suggestion
             self.api_model = cfg["AZURE_OPENAI_DEPLOYMENT"]
             api_version = cfg["OPENAI_API_VERSION"] or "2024-02-01"
             self.client = AzureOpenAI(api_key=cfg["AZURE_OPENAI_API_KEY"], azure_endpoint=cfg["AZURE_OPENAI_ENDPOINT"], api_version=api_version)
```

mspronesti (changes requested):

sweagent/agent/models.py
klieret: we could make it a `.get("...", default value)` to get the best of both worlds

mspronesti: Suggestion updated :)

klieret: Thanks! Could you check that this works? (I don't have azure credits atm)

klieret:
Since this doesn't seem to impact any other models anyway, I've gone ahead and merged it :)

klieret:
Note: We shouldn't override the `--model_name`, so I've removed `AZURE_OPENAI_DEPLOYMENT` in #127

PR #17, opened 2024-04-03
fix: use nonblocking reads to avoid unix specific apis

cpendery:
`signal.alarm` and `signal.SIGALRM` are Unix specific so a Windows user cannot start the `run.py` script. This PR changes the timeout method to use a threading timer which is cross platform.


klieret (changes requested):

sweagent/environment/utils.py
```diff
@@ -132,16 +132,16 @@ class timeout:
     def __init__(self, seconds=TIMEOUT_DURATION, error_message="Timeout"):
         self.seconds = seconds
         self.error_message = error_message
+        self.alarm = threading.Timer(self.seconds, self.handle_timeout)
```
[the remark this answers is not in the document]
klieret: I guess you could just remove the positional args in `handle_timeout`

klieret (changes requested):
More generally, on osx, this seems to raise a TimeoutError even though no such scenario should occur (not really sure at the moment)

cpendery:
> More generally, on osx, this seems to raise a TimeoutError even though no such scenario should occur (not really sure at the moment)

I realized that the previous code would throw the timeout error in the child thread causing the issue you saw, the most recent change should fix it

klieret:
I'll have a closer look in a second, just curious: With these changes, are you able to run the pipeline on Windows? Or does it immediately run into another problem?

cpendery:
> Thanks @cpendery
> 
> I'll have a closer look in a second, just curious: With these changes, are you able to run the pipeline on Windows? Or does it immediately run into another problem?

It's still running into issues. I'm working on fixing those now too. Then next one is the use of `select.select`. I can bundle them all into 1 bigger PR?

klieret:
I think it's better to create separate PRs. I'm just trying to get an overview of how difficult it is to get this running on Windows at all (because you could always install WSL and run it with that, right?)

cpendery:
@klieret it's going to need some additional work on WSL as well. The `-m` flag for bash in the container fails with the following error:
```
bash: cannot set terminal process group (-1): Inappropriate ioctl for device
bash: no job control in this shell
```

talentoscope:
os.set_blocking is not available in Windows. Just a heads up.

klieret:
@talentoscope According to the documentation it should be since python 3.5:

image

talentoscope:
Unfortunately not :(
image

cpendery:
Hi @klieret I think this is ready to merge, is there additional testing the team is waiting on?

klieret:
Hi @cpendery sorry for the late follow ups. I just tested again on osx and I get

```
DEBUG    Starting container with command: docker run -i --rm --name sweagent-swe-agent-latest-e6bd95bd78 sweagent/swe-agent:latest /bin/bash -l -m
ERROR    Unexpected container setup output: /bin/bash: cannot set terminal process group (-1): Inappropriate ioctl for device
         /bin/bash: no job control in this shell

```

but not on `main`. So somehow the timeout doesn't work on OSX at the moment :|

cpendery:
> Hi @cpendery sorry for the late follow ups. I just tested again on osx and I get
> 
> ```
> DEBUG    Starting container with command: docker run -i --rm --name sweagent-swe-agent-latest-e6bd95bd78 sweagent/swe-agent:latest /bin/bash -l -m
> ERROR    Unexpected container setup output: /bin/bash: cannot set terminal process group (-1): Inappropriate ioctl for device
>          /bin/bash: no job control in this shell
> ```
> 
> but not on `main`. So the timeout doesn't work on OSX at the moment :|

Would you mind running it again from `main`, but check the docker container's logs during execution? I've seen that error from the main branch on WSL in the container itself, so I'm curious if it's just not getting surfaced right now in OSX.

klieret:
Sorry, could you explain what you mean with "the docker container's logs during execution"? I've checked if there's any additional logs in the Docker desktop app, but they were empty. But I've tested again on `main` and on this branch, and I only see it on this branch :|

cpendery:
@klieret I managed to get access to a Mac (apple silicon) for testing and it looks like the job control issue exists even in the main branch. I think the fact you're getting the error of this branch is an improvement rather than it not getting picked up in the current implementation.

Screenshot 2024-04-12 at 10 36 39 AM

Can we remove the `-m` switch for job control if it's always failing?

https://github.com/princeton-nlp/SWE-agent/blob/fead564476940005ce2aad7135f1614b17fb785e/sweagent/environment/utils.py#L172

klieret:
Hi @cpendery , sorry for my slow responses. 

I have since also run other experiments related to `read_with_timeout` (#230) and got feedback from the rest of the team that this function was quite finicky (for example, the strategy in #230 was used previously, but apparently sometimes failed to get the full output of the command). I think the cleanest way would be something that put a flask server on the container and communicate with that rather than the whole piping business, but that would be a bit more work. In the meantime, I want to test if I can run over all of the most recent trajectories (which will probably take several hours of runtime) and reproduce the patches. The same would then go for this PR.

Also, this PR got more relevant, because you also can't use `signal` from within a thread, which might be required to use `swe-agent` together with a `flask` setup.

TL;DR: This is higher priority now, but I need time to properly test it.

klieret:
OK, noticing that there are two different things, `signal` and `select`, let's do the following:

1. The `signal` issue can be used with threads, so let's do that. This will unblock the web frontend thing and there's no danger with corrupting any outputs.
2. The `select` is more critical, but since we already have one alternative `read_with_timeout_experimental` version in, let's include that there (it's selectable with an ENV flag). This way, we remain backwards compatible, but have an easy way of continuing to test this.

cpendery:
> Hi @cpendery , sorry for my slow responses.
> 
> I have since also run other experiments related to `read_with_timeout` (#230) and got feedback from the rest of the team that this function was quite finicky (for example, the strategy in #230 was used previously, but apparently sometimes failed to get the full output of the command). I think the cleanest way would be something that put a flask server on the container and communicate with that rather than the whole piping business, but that would be a bit more work. In the meantime, I want to test if I can run over all of the most recent trajectories (which will probably take several hours of runtime) and reproduce the patches. The same would then go for this PR.
> 
> Also, this PR got more relevant, because you also can't use `signal` from within a thread, which might be required to use `swe-agent` together with a `flask` setup.
> 
> **TL;DR: This is higher priority now, but I need time to test it on a large scale.**

Question on this, are you trying to preserve a single shell session between commands? If not, you could just run `docker exec` commands into the container and wait on that output which wouldn't be flakey.

klieret:
> Question on this, are you trying to preserve a single shell session between commands? If not, you could just run docker exec commands into the container and wait on that output which wouldn't be flakey.

Yes, swe-agent runs commands in a single shell session (i.e., everything that happens in the `SWEEnv` session will be in the same shell). This is because it's running `setup` commands that might change env variables.

cpendery:
> > Question on this, are you trying to preserve a single shell session between commands? If not, you could just run docker exec commands into the container and wait on that output which wouldn't be flakey.
> 
> Yes, swe-agent runs commands in a single shell session (i.e., everything that happens in the `SWEEnv` session will be in the same shell). This is because it's running `setup` commands that might change env variables.

That makes sense, it might be overkill, but VSCode tracks when a command is completed, it's exit code, and its output using this [shell integration](https://github.com/microsoft/vscode/blob/main/src/vs/workbench/contrib/terminal/browser/media/shellIntegration-bash.sh) script.

sweagent/environment/utils.py
```diff
@@ -263,15 +250,13 @@ def _get_non_persistent_container(ctr_name: str, image_name: str) -> Tuple[subpr
         text=True,
         bufsize=1, # line buffered
     )
+    os.set_blocking(container.stdout.fileno(), False)
```
klieret, line 253: not needed

sweagent/environment/utils.py
```diff
@@ -319,15 +304,13 @@ def _get_persistent_container(ctr_name: str, image_name: str, persistent: bool =
...
```
klieret, line 307: not needed

klieret (changes requested):

klieret:
> That makes sense, it might be overkill, but VSCode tracks when a command is completed, it's exit code, and its output using this [shell integration](https://github.com/microsoft/vscode/blob/main/src/vs/workbench/contrib/terminal/browser/media/shellIntegration-bash.sh?rgh-link-date=2024-04-16T23%3A12%3A48Z) script.

Thanks, I'll take a look

klieret:
OK, maybe yet another strategy: Let's make the refactoring of the `timeout` class into the `read_with_timeout` a separate PR and merge it first (since there's less concerns about input corruptions). And that would already unblock the web server.

klieret:
>  Would you mind running it again from main, but check the docker container's logs during execution? I've seen that error from the main branch on WSL in the container itself, so I'm curious if it's just not getting surfaced right now in OSX.

Can directly reproduce this with `docker run -i --rm  sweagent/swe-agent /bin/bash -l -m`, so I think this indeed means that the error was always there, but the previous timeout never caught it. 

Interestingly, it goes away when adding the `-t` flag (`docker run -it --rm  sweagent/swe-agent /bin/bash -l -m  `)

Issue #26, opened 2024-04-03
Hello, how can I specify my openai_base_URL instead


yxl23:
Hello, how can I specify my openai_base_URL instead

klieret:
Could you elaborate?

0xdevalias:
Googling for `OPENAI_BASE_URL`, it sounds like this may commonly be used to proxy the API/similar:

- https://github.com/openai/openai-python/issues/1214
  - https://github.com/openai/openai-python#configuring-the-http-client

It sounds like it can be configured as an env var that will presumably be picked up by the underlying `openai` library.

`SWE-agent` also allows config via `keys.cfg`:

- https://github.com/princeton-nlp/SWE-agent#-setup-

Searching the repo for `keys.cfg` gives the following references:

- https://github.com/search?q=repo%3Aprinceton-nlp%2FSWE-agent%20keys.cfg&type=code
  - https://github.com/princeton-nlp/SWE-agent/blob/6c9ebf0ea8a263806b276da7ba3b1eda1f4a9475/sweagent/environment/swe_env.py#L86-L92
  - https://github.com/princeton-nlp/SWE-agent/blob/6c9ebf0ea8a263806b276da7ba3b1eda1f4a9475/sweagent/agent/models.py

From those code samples, it looks like `keys.cfg` only defines certain well-known keys that are explicitly read from that file by name; and they seem to only be API tokens/access keys, so doesn't seem like the right place to support something like `OPENAI_BASE_URL`.

So I would say that just exporting the env var in your shell environment is likely the best way to achieve this.

klieret:
> It sounds like it can be configured as an env var that will presumably be picked up by the underlying openai library.

Thanks! Then let me close this :)

PR #34, opened 2024-04-04
add TOGETHER_API_KEY to keys.cfg section of README

0xdevalias:
I noticed there is also a `TOGETHER_API_KEY` key that can be set in `keys.cfg` while I was investigating https://github.com/princeton-nlp/SWE-agent/issues/26#issuecomment-2035899278, but it wasn't mentioned in the README, so I wanted to add it:

https://github.com/princeton-nlp/SWE-agent/blob/6c9ebf0ea8a263806b276da7ba3b1eda1f4a9475/sweagent/agent/models.py#L509-L511

I also moved the `GITHUB_TOKEN` to the top of the list, since it's a required token.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,9 +49,10 @@ Read our paper for more details.
 4. Run `./setup.sh` to create the `swe-agent` docker image.
 5. Create a `keys.cfg` file at the root of this repository and fill in the following:
 ```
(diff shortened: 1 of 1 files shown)


klieret (approved):
Thanks!

Issue #36, opened 2024-04-04
Show gold patch in trajectory viewer (for SWE-bench issues)


ofirpress:
We used to have it so that on swe-bench issues you'd not only see the submitted patch but also the gold solution, we should put it back in.

PR #37, opened 2024-04-04
Create PR for changes

themrzmaster:


klieret:
Nice :) I was a little bit faster in #36 and added a little bit more code to check when it makes sense to open the PR, but I should have also just used the `GhApi` instead of handcrafting post requests. If you had opened earlier, I could've pushed on top of your commit, but I'm not sure how to include your commit now; sorry!

PR #39, opened 2024-04-04
Github action & typer CLI

msnidal:


klieret:
This seems to become a larger PR, it might make sense to open an issue and discuss first ;)

klieret:
We will almost certainly not be able to accept a PR with so many changes in one. I'd encourage you to split up your work into chunks that we can easily review (and it's not clear to me that moving to `typer` is worth it -- this would be better discussed in an issue).

zachblume:
The github action packaging is clever / great idea. Lots of people have free on-demand compute this way. Are there other example we could look at of any other open-source coding agents that have provided orchestration wrappers this way?

msnidal:
Hey guys sorry this was completely unintentional! Meant to open this against my local forked copy, was just playing around last night with making a github action - will follow up with an upstream PR if I get something working

Issue #60, opened 2024-04-04
Web interface


zachblume:
Happy to contribute here, I'm sure others are already thinking here but figured I'd write some notes about what would've made trying this out easier:

 - A locally served onboarding and output UI
 - `git clone`, run `make` spins up and opens `localhost:3000`
 - "Playground":
     - Paste GH issue (or file selector->a directory & a textbox for issue description)
     - Config chooser
 - Tail output
 - Choose to apply the diff locally or generate the PR back up to GH
 - Making this available locally vs. hosting the UI and provisioning compute for the agent (e.g. inside a Docker-layer lambda)?

klieret:
We're still discussing a bit, but if we go more towards that direction, we really should set up the web app around a flask server. Much more natural integration in the rest of the codebase that way. In particular, if we add interactivity between the user and the model (i.e., the user being able to stop the run), then using `os.system` calls is a big limitation.

zachblume:
That makes sense -- I won't be able to get anything done in the next few days, but I can plan to rework my PR to be based on a more Python/Flake-centric approach picking up whats been done in `/inspector`. 

It makes sense because as I began running this on real issues I was familiar and watching it, there were moments here I wished the agent had asked me a scope-clarifying question that I could've answered. Perhaps with a interactivity timeout (e.g. continue if not answered in 60s)?

Question for you all -- how happy are you with the state of inspector? Do you want to just add this UI to that existing server, adopt the same pattern but in a new dir like this, or start afresh in a seperate dir but with a rethought Flask pattern?

ofirpress:
Sorry I just saw this now-

The inspector will probably stay until we finish developing this new web ui, but once we have it we might deprecate the inspector.

I'm closing this issue since we'll move this discussion to #213

PR #61, opened 2024-04-05
Locally hosted web UI

zachblume:
re: https://github.com/princeton-nlp/SWE-agent/issues/60


klieret:
There's also a web-based trajectory viewer in `inspector/` that would be cool to include if we add a web interface.

More generally, though, might be worth checking Carlos/John/Ofir about whether the web UI fits into the scope of the project -- but could certainly be cool ;)

zachblume:
Happy to keep cranking on this but would prefer others thoughts first to not go down the wrong road.

```
cd app
npm i
npm run dev
```

Should give you a basic idea of the current state of things

klieret:
I hope to take a look later today or tmr morning!

ofirpress:
This has been deprecated by the new web UI. Thanks

PR #65, opened 2024-04-05
Fix searching for flag-like strings, e.g., search_file "--flag"

CodeCreator:
There is a rare issue when the agent searches for search terms starting with dashes, e.g. `search_file "--flag"`. These are interpreted as flags by the underlying grep command, which results in errors ("unrecognized flag") or in the worst case hanging since the search term is missing.

Separating flags from positional arguments with "--" should do the trick.


klieret (approved):
LGTM!

PR #72, opened 2024-04-05
refactor: Simple refactoring for clean code

YeonwooSung:
- refactor codes for clearness

### DIFF
--- a/run.py
+++ b/run.py
@@ -183,8 +183,8 @@ def should_open_pr(args, info: Dict[str, Any], *, token: str="") -> bool:
             return False
         else:
             logger.warning(
-                f"Proceeding with PR creation even though there are already commits "  # noqa: F541
-                "({commit_url_strs}) associated with the issue. Please only do this for your own repositories "
+                "Proceeding with PR creation even though there are already commits "
+                f"({commit_url_strs}) associated with the issue. Please only do this for your own repositories "
                 "or after verifying that the existing commits do not fix the issue."
             )
     return True
--- a/run.py
+++ b/run.py
@@ -183,7 +183,7 @@ def should_open_pr(args, info: Dict[str, Any], *, token: str="") -> bool:
             return False
         else:
             logger.warning(
-                f"Proceeding with PR creation even though there are already commits "
+                f"Proceeding with PR creation even though there are already commits "  # noqa: F541
                 "({commit_url_strs}) associated with the issue. Please only do this for your own repositories "
                 "or after verifying that the existing commits do not fix the issue."
             )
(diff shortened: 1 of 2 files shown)


run.py
```diff
@@ -183,7 +183,7 @@ def should_open_pr(args, info: Dict[str, Any], *, token: str="") -> bool:
             return False
         else:
             logger.warning(
-                f"Proceeding with PR creation even though there are already commits "
+                f"Proceeding with PR creation even though there are already commits "  # noqa: F541
                 "({commit_url_strs}) associated with the issue. Please only do this for your own repositories "
```
klieret, line 187: The real issue is that the next line is missing the `f` for the f-str

klieret: and if flake complains about superfluous fs, it's better to remove them than to ignore the warnings

klieret (changes requested):

run.py
YeonwooSung: Thanks for the comment. Will change it right now

sweagent/agent/models.py
```diff
@@ -690,16 +690,21 @@ def get_model(args: ModelArguments, commands: Optional[list[Command]] = None):
...
         commands = []
 
-    if args.model_name == "human":
-        return HumanModel(args, commands)
-    if args.model_name == "human_thought":
-        return HumanThoughtModel(args, commands)
-    if args.model_name == "replay":
-        return ReplayModel(args, commands)
-    elif args.model_name.startswith("gpt") or args.model_name.startswith("ft:gpt"):
+    models = {
```
klieret, line 693: honestly perhaps let's just bring that back to the original version (conflicting anyway). Shouldn't make too many changes while we don't have proper CI in place

klieret:
I think we can go ahead with this after you resolve the conflicts

klieret:
For the future it would be better to split up fixes and refactoring (and I don't think we have enough bandwidth to welcome much refactoring, especially because we don't have much CI to test it)

klieret (approved):

sweagent/agent/models.py
```diff
@@ -696,10 +696,8 @@ def get_model(args: ModelArguments, commands: Optional[list[Command]] = None):
         return HumanThoughtModel(args, commands)
     if args.model_name == "replay":
         return ReplayModel(args, commands)
-    elif args.model_name.startswith("gpt") or args.model_name.startswith("ft:gpt"):
+    elif args.model_name.startswith("gpt") or args.model_name.startswith("ft:gpt") or args.model_name.startswith("azure:gpt"):
         return OpenAIModel(args, commands)
-    elif args.model_name.startswith("claude"):
```
klieret, line 701: Why did you delete this?

klieret (changes requested):

sweagent/agent/models.py
YeonwooSung: I tried to undo the changes for the model list, and seems like I copy and pasted the code from wrong commit.. Sorry, my bad

klieret (approved):

PR #79, opened 2024-04-05
fix: bad newline getting sent on windows

cpendery:
Fixes an issue with windows where `\r\n` is getting sent to the docker image instead of `\n`. This is documented in the [subprocess](https://docs.python.org/3/library/subprocess.html) api as it replaces `\n` with `os.linesep`. With #17, this can be fully run on windows now


cpendery:
@klieret This & #17 should get everything working on windows natively now

klieret:
Oh wow, that would be a big success! Let me review this and #17 later today!

klieret (approved):

PR #80, opened 2024-04-05
style: add formatting via ruff

cpendery:
Added [ruff](https://astral.sh/ruff) to the pre-commit config. Everything else is from `ruff format`


klieret:
Let's hold back on this for now. This is also on my todo list, but let's reserve this for a later time when the rest of the team has more bandwidth and then bundle this up with some more automatic code quality improvements. I want to hold back right now as this will impact everyone's dev process.

klieret:
I deliberately also added some very basic flake8 checks rather than everything that `ruff` offers to avoid distractions. But I would be open to adding more checks right now that catch real bugs without introducing too many changes (i.e., no code quality/style things for now)

Issue #66, opened 2024-04-05
Streamlining the installation and usage process


samuela:
(posting at @ofirpress's request)

The current installation process requires a number of steps, and requires a non-trivial investment for non-contributors. For example, someone from discord recently mentioned that this process took them 2 days. 
 
Ideally swe-agent would have a single CLI command that anyone could run from any machine, regardless of whether they have cloned the repo or downloaded any dependencies:
```
$ swe-agent \
  --github-issue=https://github.com/princeton-nlp/SWE-agent/issues/57 \
  --model=gpt4 \
  --openai_api_key=... \
  --cost-limit=2.00
```
  
# Design possibilities
Note that none of these solutions are mutually exclusive.
 
## Docker container
**Usage:** `docker run princeton-nlp/swe-agent ...`
 
In this solution, we would publish a Docker container with `CMD` set to `python run.py`.
 
Considerations:
* Would require the user to have Docker installed before running the command.
* Would involve docker-in-docker, which is do-able but brings its own subtleties.
 
## Nix flake
**Usage:** `nix run github:princeton-nlp/SWE-agent`
 
In this solution, we would package swe-agent as a [nix](https://nixos.org/) [flake](https://zero-to-nix.com/concepts/flakes). This has the advantage of increasing reproducibility and obviating the need for the current `setup.py`/`pyproject.toml` situation.
 
Considerations:
* Would require the user to have Nix, and possibly Docker, installed before running the command. While nix has packaged the Docker CLI, the Docker daemon will likely require a native installation on non-NixOS operating systems.
* Pro: would not require docker-in-docker
* Pro: dependency pulling and caching would be handled automatically by Nix
 
## curl-to-shell
**Usage:** `curl https://github.com/princeton-nlp/SWE-agent/blob/main/run.sh | sh`
 
Here we would be following in the footsteps of [rustup](https://rustup.rs/), [the Nix installer](https://nixos.org/download/), and friends.
 
Considerations
* Pro: would involve minimal changes to the existing packaging of SWE-agent
* Pro: would require effectively no pre-requisites
* Con: would require prompting for sudo to install Docker on Linux, and may not be possible on macOS
* Con: would require implementing logic to query, download, and cache this repo and its dependencies on the user's local machine. This would effectively be a reimplementation of package management logic that already exists in Nix/Docker.
[...]

PR #81, opened 2024-04-05
Containerize application

### LINKED ISSUE #28: Please create a docker build so that the local conda install is not needed
The whole point of docker is that images can be run as released, without the user having to rebuild it every time. 
If the API keys are the issue, **config files, environmental variables and docker secrets** can be used to circumvent having to bake api keys into the build.

klieret:
Part of a solution to #66 and might help people running Windows as well.

Current way of running this:

```bash
docker build -t swe-agent-run:latest .  && \
docker run -it -v /var/run/docker.sock:/var/run/docker.sock --rm swe-agent-run:latest  \
python3 run.py --model_name gpt4 \
  --data_path  https://github.com/klieret/swe-agent-test-repo/issues/1 \
  --config_file config/default_from_url.yaml
```

Closes #28 

## Tasks

* [x] ~~Add `run.py` as default entry point~~ let's not do this. we might have more executables in here in the future and this doesn't really make things simpler
* [ ] Push working version to dockerhub
* [ ] Push swe-agent container to dockerhub
* [ ] Add instruction to readme

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,16 @@
+FROM python:3.9
+
+# Set the working directory
+WORKDIR /app
+
+# Install Python dependencies
+RUN pip install anthropic config datasets docker gymnasium numpy openai pandas rich ruamel.yaml swebench tenacity unidiff simple-parsing together ollama
+
+# Install Docker CLI using the official Docker installation script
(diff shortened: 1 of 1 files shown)


klieret:
Merging this right now because should have it in the repo to have the same version then on dockerhub

PR #83, opened 2024-04-05
feat: support languages beyond python

### LINKED ISSUE #53: Support agent working on codebases outside of python
It would be great to have the source code language be configurable beyond just python. It would involve tweaking some of the prompts 

https://github.com/princeton-nlp/SWE-agent/blob/3d8a6c44f16f190735c774676fea80315cd84da5/config/default.yaml#L38

and also changing the shell scripts to customize the linter that is run for different languages like `golangci-lint` for Go, but I'm not sure limitations there might be.

https://github.com/princeton-nlp/SWE-agent/blob/3d8a6c44f16f190735c774676fea80315cd84da5/config/commands/edit_linting.sh#L72

cpendery:
Closes #53

### DIFF
--- a/docker/swe.Dockerfile
+++ b/docker/swe.Dockerfile
@@ -5,6 +5,7 @@ ARG MINICONDA_URL
 # Install third party tools
 RUN apt-get update && \
     apt-get install -y bash gcc git jq wget g++ make && \
+    apt-get install -y nodejs npm && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/*
 
--- a/config/commands/_split_string.py
+++ b/config/commands/_split_string.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 import sys
+import os
 
 def print_flake8_output(input_string, show_line_numbers=False):
     for value in input_string.split("\n"):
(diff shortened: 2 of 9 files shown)


zachblume:
```
INFO     💭 THOUGHT (primary)                                                                                                                                                                                          
         It seems there was an error with the edit command, possibly due to a syntax issue in the environment's processing of the command rather than the edit itself. The edit proposed was correct according to the  
         file's context, so I will attempt the edit command again.                                                                                                                                                     
                                                                                                                                                                                                                       
         DISCUSSION                                                                                                                                                                                                    
         I will re-issue the edit command with the same content, as the error message seems to be unrelated to the actual edit content.                                                                                
                                                                                                                                                                                                                       
                                                                                                                                                                                                                       
INFO     🎬 ACTION (primary)                                                                                                                                                                                           
         edit 591:591                                                                                                                                                                                                  
                     valueGetter: ({ row }) => row?.phone_numbers?.phone_number,                                                                                                                                       
         end_of_edit                                                                                                                                                                                                   
                                                                                                                                                                                                                       
INFO     Saved trajectory to trajectories/zachblume/gpt4__zachblume__raisemore__default_from_url_w_lang__t-0.00__p-0.95__c-3.00__install-1/zachblume__raisemore-i289.traj                                              
INFO     🤖 MODEL INPUT                                                                                                                                                                                                
         Your proposed edit has introduced new syntax error(s). Please understand the fixes and retry your edit commmand.                                                                                              
                                                                                                                                                                                                                       
         ERRORS:                                                                                                                                                                                                       
           File "/root/commands/_split_string", line 27                                                                                                                                                                
             match language:                                                                                                                                                                                           
                   ^                                                                                                                                                                                                   
         SyntaxError: invalid syntax   
```

I pulled this PR, merged main, and ran it on a javascript project and got this

zachblume:
https://docs.sweep.dev/blogs/super-linter

This is a good discussion of cross-language linting in a code generation agent

docker/swe.Dockerfile
```diff
@@ -5,6 +5,7 @@ ARG MINICONDA_URL
 # Install third party tools
 RUN apt-get update && \
     apt-get install -y bash gcc git jq wget g++ make && \
+    apt-get install -y nodejs npm && \
```
jvmncs, line 8: seems like this will bloat the image as more languages are added according to the thrust of [this section](https://github.com/princeton-nlp/SWE-agent/pull/83/files#diff-404c46b336280a67ec4b902a13f28d1cb24fdc2f17f608bc586eeedfada91ca7R27). I'm no docker expert but I wonder if there's a way to parameterize this dockerfile somehow

cpendery: Some of the initial setup of the container still passes in manual install commands, so we could technically move this there if container bloat becomes an issue

PR #84, opened 2024-04-05
Add github action for swe agent 

ezzcodeezzlife:
This is an easily sharable action that can be added to other repos to solve newly created issues. 

While I believe this feature could be a valuable addition, I understand that it might not be the perfect solution. Therefore, I am open to feedback and suggestions for improvement.

### DIFF
--- a/.github/workflows/swe-agent.yml
+++ b/.github/workflows/swe-agent.yml
@@ -1,0 +1,49 @@
+name: SWE-agent
+
+on:
+  issues:
+    types: [opened]
+
+jobs:
+  setup_job:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Setup Miniconda
+        uses: conda-incubator/setup-miniconda@v2
+        with:
+          miniconda-version: "latest"
+          architecture: x64
+      - name: Clone SWE-agent
+        run: |
+          git clone https://github.com/princeton-nlp/SWE-agent
+        shell: bash
+      - name: Create environment
+        run: |
+          cd SWE-agent
+          conda env create -f environment.yml
+        shell: bash
+      - name: Run setup script and create keys.cfg
+        run: |
+          source $(conda info --base)/etc/profile.d/conda.sh
+          conda activate swe-agent
+          cd SWE-agent
+          chmod +x setup.sh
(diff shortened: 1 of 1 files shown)


zachblume:
```
  workflow_dispatch:
    inputs:
      issueNumber:
        description: 'Issue Number'
        required: true
        type: integer
```

Let people kick it off for past issues

klieret:
This is potentially dangerous: Let's say you install this in your repo and add your API keys. Now I write a script that opens 1000 trash issues. You'll loose 2000$ in OpenAI credits.

klieret:
But there's also a less glorious but potentially more useful use for us: Running unittests (possibly via the `run_replay` script so it doesn't actually cost us anything) ;)  So as a developer I'm very happy if this works!

dsernst:
> Thanks for creating this!
> 
> 
> 
> However: **⚠️ This is potentially dangerous: ⚠️** Let's say you install this in your repo and add your API keys. Now I write a script that opens 1000 duplicate issues. You'll loose 2000$ in OpenAI credits.
> 
> 
> 
> Any ideas of how to avoid something like this? 


Can start by enabling it to auto-run only for issue openers who already have commit privileges. The power users that want to improve their own repos.

For other users (visitors who open  ew bug reports etc), can require admin approval before calling the LLM API, or whitelisted users, or eg limiting to accept their first X. Just some ideas.

klieret (changes requested):
We can't merge it like this. Any such solution should also have a mechanism to "vet" the issue before attempting anything (similar to the logic that we have to determine whether we open a PR with a change or not, but before we even start querying a LLM). One use case of this snippet (with some modifications) that I do see is as a basis of a testing pipeline, but this would be better as a separate PR.

klieret:
I will convert this to a draft for now.

klieret:
I have merged #128 which contains a github action that runs a replay of an existing trajectory. I don't think the github action in this PR would have run, so I'm closing this one (because it makes more sense to base any new attempt to do this on the working example with the tests). But as I said, I think there's some more work to be done before we can have a GHA  that attempts something like you described (most notably the security features).

PR #90, opened 2024-04-06
Add run via docker instructions to readme

klieret:
* [x] Show how to get keys without cloned repo
* [ ] Link to subsections


klieret (commented):
Add note about API keys with docker usage

Issue #64, opened 2024-04-05
Using LiteLLM to support more models


Greatz08:
This project is pretty great BUT we need more options to use different LLM's.You don't have to worry about creating a solution which supports 100+ LLM easily as LiteLLM is another foss project which is capable of doing this task for you.

Project LiteLLM link - https://github.com/BerriAI/litellm

Adding LiteLLM will be big win for the project as many will be easily able to use many more LLM easily which everyone wants and project will require 3 major parameters from user like base url,model name,api key that's all and with open ai api general structure it can query and give back result for the query.Many big projects have started adding support for this project in there project to make things advanced in easier way so study it and after that if you have any query you can ask them they are pretty responsive plus if u want to know more about my personal experience of using it with other great projects like flowise then I can tell you that too .

klieret:
Sounds nice, I agree that this makes sense. However, this would need  some amount of refactoring (cost tracking for example needs to be set up [differently](https://docs.litellm.ai/docs/#track-costs-usage-latency-for-streaming)).

ofirpress:
This is a super low priority issue right now. Most LMs would not really be able to achieve good performance on SWE-agent + SWE-bench... so this would be a waste of time right now + be hard to maintain probably?

erkinalp:
@ofirpress Not only that, there is another coding assistant that offers LiteLLM support: https://github.com/OpenDevin/
It is also under MIT licence.

HakaishinShwet:
@ofirpress Litellm make things easy for both devs and user to test and try 100+ LLM easily. Implementing Litellm right now maybe little challenging but in my opinion will make things much more easy for future rather than implementing later(with low priority mindset) when projects become much more complex.Benefit is for both and that is why many projects are implementing too.
One more thing i would like to add in this is that do not underestimate different closed source model and open source models compared to open ai gpt variants because many have equalized or outperformed them and will keep on outperforming them in future maybe in one domain/segment like just coding or sql(specific task based open source model) or in generalized way so i believe in giving users option to test whatever they want and in easiest way possible so that they can show you more interesting stuff then what u can imagine someday.
Rest i leave this matter to team All the best :-))

Aaronminer1:
add support for LM-studio so we can test with local opensource models this should be pretty easy since it uses the openai library.

xiechengmude:
Supporting local model with litellm or vllm is much more necessary to boost the project with open sources power.

elsatch:
[...]

PR #108, opened 2024-04-06
feat: add groq api and model options

gluneau:
#### What does this implement

 - Add Groq API and model options
 
 get free API key here: https://console.groq.com/keys
 
use as usual by naming the groq model in the command:
 ```
 python run.py --model_name groq-mixtral8x7b \                
  --data_path https://github.com/pvlib/pvlib-python/issues/1603 \
  --config_file config/default.yaml
 ```


sweagent/agent/models.py
```diff
@@ -272,6 +272,80 @@ def query(self, history: list[dict[str, str]]) -> str:
...
+            for entry in history
+        ]
+
+    @retry(
+        wait=wait_random_exponential(min=1, max=15),
+        reraise=True,
+        stop=stop_after_attempt(3),
+        retry=retry_if_not_exception_type((CostLimitExceededError, RuntimeError)),
+    )
+    def query(self, history: list[dict[str, str]]) -> str:
```
klieret, line 329: How similar is this and `history_to_messages` to the `OpenaAIModel` class? Because if that's just exactly the same, let's just subclass from there to cut down on duplicity.

klieret:
Note: #108 and #134 seem to have the same goal.

sweagent/agent/models.py
gluneau: It is the same since I'm using the OpenAI() client because Groq is design to be compatible with OpenAi.
 https://console.groq.com/docs/openai

crjaensch:
> Note: #108 and #134 seem to have the same goal.

I agree. I have no particular preferences whether you want to approve #134 or my PR. In fact, PR #134 seems to have a better separation of concern than the minimal surgical changes proposed by my PR.
However, I would recommend that PR #134 follows the model naming convention that you introduced for example for Azure or Ollama models. 
That means, a Groq model like Mixtral 8x7B should be prefixed as groq:model_name rather than groq_model_name as proposed in PR #134.

asapsav:
hi! just a couple of adds:
1) in `class GroqModel(BaseModel):`  llama2 should be changed to "llama3-70b-8192"
2)  after a couple of agent iterations (using groq) an "Exit due to cost limit" error appears, which does not make sence cause groq is 0$

klieret:
Apologies letting this PR hang for such a long time. Part of the reason is that `models.py` is just growing and growing and we should instead just support `litellm` (potentially leaving claude and gpt as before to make sure the paper results are still reproducible, unless we check it's 100% of a drop-in replacement). The other reason is that so far only the largest models (gpt, claude) were good enough to make sense for swe-agent, so that pushed the support for other models down the list of priorities. Let me get back to this later this week!

klieret:
Closing because we should use litellm to support more models (#64 )

PR #118, opened 2024-04-07
add npx and openapi base url and model name keys - resolves #64

### LINKED ISSUE #64: Using LiteLLM to support more models

bvandorf:
#### Any other comments?

🧡 Thanks for contributing!

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -32,7 +32,6 @@ setuptools.setup(
         'datasets',
         'docker',
         'gymnasium',
-        'npx',
         'numpy',
         'openai>=1.0',
         'pandas',
--- a/setup.py
+++ b/setup.py
@@ -32,6 +32,7 @@ setuptools.setup(
         'datasets',
         'docker',
         'gymnasium',
+        'npx',
         'numpy',
         'openai>=1.0',
         'pandas',
--- a/sweagent/agent/models.py
+++ b/sweagent/agent/models.py
@@ -227,9 +227,8 @@ class OpenAIModel(BaseModel):
             self.api_model = cfg["AZURE_OPENAI_DEPLOYMENT"]
             self.client = AzureOpenAI(api_key=cfg["AZURE_OPENAI_API_KEY"], azure_endpoint=cfg["AZURE_OPENAI_ENDPOINT"], api_version=cfg.get("AZURE_OPENAI_API_VERSION", "2024-02-01"))
         else:
-            self.api_base_url = cfg.get("OPENAI_BASE_URL", None)
-            self.api_model = cfg.get("OPENAI_MODEL_NAME", None)
-            self.client = OpenAI(api_key=cfg["OPENAI_API_KEY"], base_url=self.api_base_url)
+            api_base_url: Optional[str] = cfg.get("OPENAI_API_BASE_URL", None)
+            self.client = OpenAI(api_key=cfg["OPENAI_API_KEY"], base_url=api_base_url)
 
     def history_to_messages(
         self, history: list[dict[str, str]], is_demonstration: bool = False
--- a/sweagent/agent/models.py
+++ b/sweagent/agent/models.py
@@ -227,7 +227,9 @@ class OpenAIModel(BaseModel):
             self.api_model = cfg["AZURE_OPENAI_DEPLOYMENT"]
             self.client = AzureOpenAI(api_key=cfg["AZURE_OPENAI_API_KEY"], azure_endpoint=cfg["AZURE_OPENAI_ENDPOINT"], api_version=cfg.get("AZURE_OPENAI_API_VERSION", "2024-02-01"))
         else:
-            self.client = OpenAI(api_key=cfg["OPENAI_API_KEY"])
+            self.api_base_url = cfg.get("OPENAI_BASE_URL", None)
+            self.api_model = cfg.get("OPENAI_MODEL_NAME", None)
+            self.client = OpenAI(api_key=cfg["OPENAI_API_KEY"], base_url=self.api_base_url)
 
     def history_to_messages(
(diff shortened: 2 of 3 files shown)


mspronesti (commented):
Does the token counting work when using a self-hosted model?

sweagent/agent/models.py
```diff
@@ -227,7 +227,9 @@ def __init__(self, args: ModelArguments, commands: list[Command]):
             self.api_model = cfg["AZURE_OPENAI_DEPLOYMENT"]
             self.client = AzureOpenAI(api_key=cfg["AZURE_OPENAI_API_KEY"], azure_endpoint=cfg["AZURE_OPENAI_ENDPOINT"], api_version=cfg.get("AZURE_OPENAI_API_VERSION", "2024-02-01"))
         else:
-            self.client = OpenAI(api_key=cfg["OPENAI_API_KEY"])
+            self.api_base_url = cfg.get("OPENAI_BASE_URL", None)
+            self.api_model = cfg.get("OPENAI_MODEL_NAME", None)
```
klieret, line 231: This overrides the setting of the model name via `self.args.model_name`. It's actually also incorrectly done with the `AZURE_OPENAI_DEPLOYMENT`. I will separately clean this up. `self._api_model` is set in `BaseModel.__init__`

sweagent/agent/models.py
```diff
@@ -227,7 +227,9 @@ def __init__(self, args: ModelArguments, commands: list[Command]):
             self.api_model = cfg["AZURE_OPENAI_DEPLOYMENT"]
             self.client = AzureOpenAI(api_key=cfg["AZURE_OPENAI_API_KEY"], azure_endpoint=cfg["AZURE_OPENAI_ENDPOINT"], api_version=cfg.get("AZURE_OPENAI_API_VERSION", "2024-02-01"))
         else:
-            self.client = OpenAI(api_key=cfg["OPENAI_API_KEY"])
+            self.api_base_url = cfg.get("OPENAI_BASE_URL", None)
```
klieret, line 230: seems like this should be a local variable, right? Or why do you set it as an attribute?

klieret (changes requested):
Thanks, I think we can accept this once the comments are addressed :)

sweagent/agent/models.py
bvandorf: well if you look at the open ai code you actually can set this as an environment variable, but it is unintuitive when using it in this project since the existing implementation of settings and variables are in the keys.cfg file

sweagent/agent/models.py
bvandorf: I did go back in forth myself with using the AZURE_OPENAI_DEPLOYMENT but in the end decided since it was labeled azure openai it was specific to azure along with its endpoint setting, so a separate one was created for the plain openai that goes along with the normal openai api key setting.

as far as overriding the model name that was the point so if you set the settings to a local llm 
OPENAI_BASE_URL: 'http://localhost:1234/v1'
OPENAI_MODEL_NAME: 'TheBloke/stablelm-zephyr-3b-GGUF/stablelm-zephyr-3b.Q4_K_S.gguf'



LM Studio Server Console Example
```
# Load a model, start the server, and run this example in your terminal
# Choose between streaming and non-streaming mode by setting the "stream" field
curl http://localhost:1234/v1/chat/completions \
-H "Content-Type: application/json" \
-d '{ 
  "model": "TheBloke/stablelm-zephyr-3b-GGUF/stablelm-zephyr-3b.Q4_K_S.gguf",
  "messages": [ 
    { "role": "system", "content": "Always answer in rhymes." },
    { "role": "user", "content": "Introduce yourself." }
  ], 
  "temperature": 0.7, 
  "max_tokens": -1,
  "stream": true
}'
```

bvandorf:
> Does the token counting work when using a self-hosted model?

it seems to work ok

bvandorf:
> #### Reference Issues/PRs
[https://github.com/princeton-nlp/SWE-agent/issues/64](url)
> #### What does this implement/fix? Explain your changes.
adds abilities to use local llm models that support the openai endpoints

sweagent/agent/models.py
klieret: No, I mean that it should be `api_base_url = ...` instead of `self.api_base_url`

sweagent/agent/models.py
klieret: No, I mean that the model name should be specified as a command line option (`--model_name`), not in `keys.cfg`.

klieret:
If you're ok with it, I can push a commit to address my two comments on this branch

bvandorf:
sure not a problem with me.

sweagent/agent/models.py
bvandorf: o yeah I see I was not looking at the command line argument also

setup.py
```diff
@@ -32,6 +32,7 @@
         'datasets',
         'docker',
         'gymnasium',
+        'npx',
```
klieret, line 35: I don't understand why you would add this, I'll remove it

PR #127, opened 2024-04-07
Remove azure override of model name

klieret:
This is a fixup to #16. We shouldn't simply override the `--model_name` option from the CLI with something from `keys.cfg`

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,7 +119,6 @@ ANTHROPIC_API_KEY: 'Anthropic API Key Here if using Anthropic Model (optional)'
 TOGETHER_API_KEY: 'Together API Key Here if using Together Model (optional)'
 AZURE_OPENAI_API_KEY: 'Azure OpenAI API Key Here if using Azure OpenAI Model (optional)'
 AZURE_OPENAI_ENDPOINT: 'Azure OpenAI Endpoint Here if using Azure OpenAI Model (optional)'
-AZURE_OPENAI_DEPLOYMENT: 'Azure OpenAI Deployment Here if using Azure OpenAI Model (optional)'
 AZURE_OPENAI_API_VERSION: 'Azure OpenAI API Version Here if using Azure OpenAI Model (optional)'
 OPENAI_API_BASE_URL: 'LLM base URL here if using Local or alternative api Endpoint (optional)'
 ```  
--- a/sweagent/agent/models.py
+++ b/sweagent/agent/models.py
@@ -224,7 +224,6 @@ class OpenAIModel(BaseModel):
         # Set OpenAI key
         cfg = config.Config(os.path.join(os.getcwd(), "keys.cfg"))
         if self.args.model_name.startswith("azure"):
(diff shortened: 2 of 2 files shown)


cpendery:
> This is a fixup to #16. We shouldn't simply override the `--model_name` option from the CLI with something from `keys.cfg`

This will cause some issues because **-** is forbidden from azure deployment names, which is what the model field is referencing. You can't use a specific model by name with azure openai

klieret:
Ah, I might have been too hasty here. Could you give me an example?

I guess it probably should've been something like `api_model = args.model_name.remove_prefix("[azure](azure:gpt)-")`, right?

cpendery:
> Ah, I might have been too hasty here. Could you give me an example?
> 
> I guess it probably should've been something like `api_model = args.model_name.remove_prefix("azure:gpt-")`, right?

From the [docs](https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/create-resource?pivots=web-portal#deploy-a-model), you create a deployment which represents a specific model in Azure OpenAI. The only restriction is the name has to match the regex `/[A-z0-9\-_]{2,64}/g`.

Maybe we can remove `AZURE_OPENAI_ENDPOINT` and set the model name to be something like this following setting the client
```python
self.api_model = self.api_model.replace(".", "")
```
https://github.com/princeton-nlp/SWE-agent/blob/7b5ab3df22712778bfd22726fa93f0dd247ba185/sweagent/agent/models.py#L227

In the Azure UI, they just use the model names without any periods, and users can always rename their deployments to match the model name

klieret:
Sorry if I'm being slow, but then why not just call run `--model_name=azure:gpt-whateveryounameit` when you deploy your azure model as `gpt-whateveryounameit`?

cpendery:
> Sorry if I'm being slow, but then why not just run `--model_name=azure:gpt-whateveryounameit` when you deploy your azure model as `gpt-whateveryounameit`? Like why do you assume that the model name would need to have a `.` in it?

Oh that makes sense, but how would we know which price to use for their estimate?

klieret:
Ahhh, you're right, that's a problem.

cpendery:
There is a way to use [azure-mgmt-cognitiveservices](https://pypi.org/project/azure-mgmt-cognitiveservices/) to get the model behind a deployment, but the user would then have to select a subscription id for that, which is another setting & dependency I don't think would be good to add

klieret:
OK, so in your previous version, you relied on `--model_name` to set the right `self.model_metadata`, based on `args.model_name` but then used a separate endpoint (which can be whatever).

cpendery:
I'll see if I can iterate on it and create a PR with something better. This [SO](https://stackoverflow.com/questions/77319962/is-there-a-way-via-api-to-know-what-ai-model-the-specified-azure-open-ai-endpoin) looks promising

klieret:
Nah, let's just roll back to your previous version ;)

klieret:
I reverted it. Thanks for all of your feedback @cpendery ❤️

PR #134, opened 2024-04-08
Minimal changes to support models of Groq Cloud

crjaensch:
GroqCloud offers one of the fastest inference engines available today with 500-700 tokens/sec.
They host well-known OSS models, such as Mixtral 8x7B, and support an API that is 100% compatible with the OpenAI API.
I made minimal changes to __model.py__ to support calling that API.

In addition, I created two files that show how to call the CLI for both GroqCloud models as well as local Ollama models. 

#### Any other comments?

🧡 Thanks for contributing!


klieret:
PR with same goal: #108

PR #138, opened 2024-04-08
Docs: add azure model_name arg usage

masa5555:
#### Any other comments?

🧡 Thanks for contributing!


klieret:
This now applies to multiple models. Will think of a more holistic way to solve this with #224

PR #175, opened 2024-04-10
Typing fixes

klieret:


klieret:
Doesn't work as simple as this because `simple_parsing` inspects the types.

PR #182, opened 2024-04-11
feat(agents): support Azure deployment with `--model`

mspronesti:
#### What does this implement/fix? 
This PR removes the unnecessary environment variable `AZURE_OPENAI_DEPLOYMENT` and parses the model passed to the container/script instead.

Since the prefix of `--model` is used to discriminate whether to use OpenAI, Azure OpenAI, Ollama, Claude etc., the deployment name shouldn't be specified again  in `keys.cfg`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,6 +161,19 @@ python run.py --model_name gpt4 \
 * See the [`trajectories/`](trajectories) folder for details about the output of `run.py`.
 
 <details>
+<summary> Azure OpenAI support</summary>
+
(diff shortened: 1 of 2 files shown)


klieret:
See https://github.com/princeton-nlp/SWE-agent/pull/127

mspronesti:
I closed it after realizing I was testing it with a deployment name that was "a lack valid name" :/ 
I find specifying twice the model a little inconvenient though.

PR #207, opened 2024-04-12
feat: Amazon Bedrock support (Claude models)

JGalego:
#### Reference Issues/PRs

https://github.com/princeton-nlp/SWE-agent/issues/27

This PR fixes https://github.com/princeton-nlp/SWE-agent/issues/27 (Claude models only).
It adds a `BedrockModel` class built on top of `anthropic[bedrock]`.

```bash
docker run --rm -it -v /var/run/docker.sock:/var/run/docker.sock \
    $(env | grep AWS_ | xargs printf ' -e %s') \
    sweagent/swe-agent-run:latest \
    python run.py --image_name=sweagent/swe-agent:latest \
    --model_name bedrock:anthropic.claude-3-sonnet-20240229-v1:0 \
    --data_path https://github.com/pvlib/pvlib-python/issues/1603 \
    --config_file config/default_from_url.yaml  --skip_existing=False
```


sweagent/agent/models.py
```diff
@@ -425,6 +428,157 @@ def query(self, history: list[dict[str, str]]) -> str:
...
+
+        model_provider = self.api_model.split('.')[0]
+        if model_provider == "anthropic":
+            # Note: this assumes AWS credentials are already configured.
+            # https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html
+            self.api = AnthropicBedrock()
+        else:
+            raise NotImplementedError
+
+    def history_to_messages(
```
klieret, line 473: How specific is this code and that of `query` to this class? Can't we just inherit most of this from the anthropic model? Really should cut down on duplicity.

JGalego: Thanks for reviewing this, @klieret. I was more worried about future proofing the changes than about code duplication. Even though this PR only adds support for Claude models, the idea is to prepare the class for adding other models and model providers.

```bash
def history_to_messages(...) -> list[dict[str, str]]:
    (...)
    model_provider = self.api_model.split('.')[0]
    if model_provider == "anthropic":
        # Covered by this PR
        # Adapted from the AnthropicModel implementation
    elif model_provider == "cohere":
        # Not supported
    elif model_provider == "meta":
        # Idem
    (...)
    else
        raise NoImplementedError(f"{self.api_model} is not supported!")
```

For `anthropic` models, the code is the same because you can just replace `Anthropic` with `AnthropicBedrock`, but this may not always be the case. Plus, models have different IDs e.g. `claude-3-sonnet-20240229` (1P) vs `anthropic.claude-3-sonnet-20240229-v1:0` (Bedrock) and can different lifecycles.

klieret: I understand, inheritance might indeed not be the best way then. Still, couldn't we extract some of the shared code for anthropic-style interfaces into e.g., a function that is shared between `Anthropic` and `AnthropicBedrock`?

klieret: But hoping to get this merged this week :)

JGalego: The only difference is the model name (*see images*), so it's feasible. Just have to split the `api_model` further. I'll do that in a sec.

**Anthropic**

image

**Bedrock**

image

JGalego: @klieret Changes implemented ✅ 1/ refactored `AnthropicModel` and `BedrockModel` to avoid code duplication (created methods `anthropic_history_to_messages` and `anthropic_query`), 2/ added more informative error messages (incl. model provider validation) to `BedrockModel` and 3/ added [Claude 3 Opus](https://aws.amazon.com/blogs/aws/anthropics-claude-3-opus-model-on-amazon-bedrock/)

klieret (approved):
Looking good! Thanks a lot for putting in the time! Lmk when it's ready to merge!

JGalego:
> Looking good! Thanks a lot for putting in the time! Lmk when it's ready to merge!

Retesting since the merge from `main` introduced some bugs (missing `rich_argparse` pkg when using the `swe-agent-run` image)

JGalego:
`claude-2` is now an alias for `claude-2.0` (see https://github.com/anthropics/anthropic-sdk-python/issues/255), made the change in `AnthropicModel`. @klieret ready to go 🤘

klieret (approved):

klieret:
Wait, one last thing: So when people specified `--model claude-2` or `--model claude`, it picked `claude-2.1`, right? Then I think we should actually keep this behavior for now (this would both be in line with the anthropic issue and for reproducibility of the research results).

klieret (changes requested):
will merge once we settle on the claude-2.0 issue

JGalego:
> Wait, one last thing: So when people specified `--model claude-2` or `--model claude`, it picked `claude-2.1`, right? Then I think we should actually keep this behavior for now (this would both be in line with the anthropic issue and for reproducibility of the research results).

I'll change the alias name (`claude` --> `claude-2`) and the target (`claude-2.0` --> `claude-2.1`)

klieret (approved):

PR #214, opened 2024-04-13
improve patch print message

klieret:
image


ofirpress:
LGTM! Thank you @klieret

ofirpress:
question- is this the _last_ thing the system prints? I think it should be, that way it can't be missed.

klieret:
> question- is this the last thing the system prints? I think it should be, that way it can't be missed.
yup

Issue #149, opened 2024-04-09
Speed up communication with docker container bash


ofirpress:
### Describe the bug

@klieret brought this up and I wanted to open an issue about this.

When running a replay or when using the 'human' model, the agent seems to run slow, much slower than it should. Are we missing something here? Is there some inefficiency we could improve on? 

### Steps/commands/code to Reproduce

Run a replay or open a session with the --model set to human

### Error message/results

The agent seems to run slow

### System Information

mac m2

### Checklist

- [X] I have copied the full command/code that I ran
- [X] I have copied the **full** log file/error message that was the result (if applicable)
- [X] I have enclosed code/log messages in triple backticks ([docs](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#quoting-code))

klieret:
This doesn't affect just replay/human model, this certainly affects all runs.

Here's a small snippet to collect information:

```python
_LAST_TIME = datetime.datetime.now()


def debug_time():
    """Display time delta at line number"""
    cf = currentframe()
    line_no = cf.f_back.f_lineno
    global _LAST_TIME
    time_delta = datetime.datetime.now() - _LAST_TIME
    _LAST_TIME = datetime.datetime.now()
    logger.debug(f"Time: {time_delta} at line {line_no}") 
```

then just add `debug_time()` in the code at various places.

klieret:
```
INFO     🎬 ACTION (primary)
         ls
DEBUG    Time: 0:00:00.003451 at line 674 (name='reset before postproc')
DEBUG    Time: 0:00:00.000472 at line 677 (name='guard multiline')
DEBUG    Time: 0:00:01.638950 at line 692 (name='got observation')
INFO     Saved trajectory to trajectories/fuchur/human__klieret__swe-agent-test-repo__default_from_url__t-0.00__p-0.95__c-3.00__install-1/klieret__swe-agent-test-repo-i1.traj
DEBUG    Time: 0:00:00.007787 at line 706 (name='traj saved')
DEBUG    Time: 0:00:00.001074 at line 667 (name='')
DEBUG    Time: 0:00:01.698880 at line 669 (name='state cmd')
DEBUG    Time: 0:00:00.003122 at line 390 (name='')
DEBUG    Time: 0:00:00.002390 at line 411 (name='')
INFO     🤖 MODEL INPUT
         README.md
         tests

         (Open file: n/a)
[...]

PR #230, opened 2024-04-15
Tests to use fast experimental communication strategy

klieret:
Also see #149


klieret:
This PR reduces the test time from 14min to 8min

PR #234, opened 2024-04-16
Add list of available models

### LINKED ISSUE #224: List all available models
Ideally, this should be e.g., in the `--help` text and be fetched based on `models.py`, so that it can't go out of date.

hminsky2002:
#### Reference Issues/PRs

Fixes #224 

This PR adds a list of the current available models to the --help doc in the command line. 
Modifies:
`sweagent/agent/models.py`
`sweagent/__init__.py`
`run.py` 
#### Any other comments?
I'm very open to refactoring this commit/making whatever changes are necessary to fit in the style guidelines, or if there's a better approach to solving this problem that someone else sees!
🧡 Thanks for contributing!

### DIFF
--- a/run.py
+++ b/run.py
@@ -32,13 +32,17 @@ from sweagent import (
     ModelArguments,
     SWEEnv,
     get_data_path_name,
+    get_model_names
 )
 from swebench import KEY_INSTANCE_ID, KEY_MODEL, KEY_PREDICTION
 from unidiff import PatchSet
 
 from sweagent.environment.utils import InvalidGithubURL, get_associated_commit_urls, get_gh_issue_data, parse_gh_issue_url
 
-__doc__: str = """ Run inference. Usage examples:
+# Get up-to-date list of supported models
+formatted_model_names = " | ".join(get_model_names())
+
+__doc__: str = f""" Run inference. Usage examples:
 
 ```bash
 # Run over a github issue:
@@ -46,6 +50,8 @@ python run.py --model_name "gpt4" --data_path "https://github.com/pvlib/pvlib-py
 # Apply a patch in a local repository to an issue specified as Markdown file and run a custom installer script in the container
 python run.py --model_name "gpt4" --data_path "/path/to/my_issue.md" --repo_path "/path/to/my/local/repo" --environment_setup "/path/to/setup.sh" --config_file "config/default_from_url.yaml" --apply_patch_locally
 ```
+Available Models:\n
+{formatted_model_names}
 """
 
 handler = RichHandler(show_time=False, show_path=False)
--- a/sweagent/__init__.py
+++ b/sweagent/__init__.py
@@ -7,6 +7,7 @@ from sweagent.agent.agents import (
 
 from sweagent.agent.models import (
     ModelArguments,
+    get_model_names
 )
 
 from sweagent.environment.swe_env import (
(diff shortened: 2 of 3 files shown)


klieret:
Thanks! I'll deal with the merge conflict and will push some smaller changes on top :)

hminsky2002:
> Thanks! I'll deal with the merge conflict and will push some smaller changes on top :)

Woohoo! Happy to do so, I'll keep my eyes peeled for any other issues that I could help with :)

klieret:
One of the problems is that some models require a certain prefix that they strip (see for example the upcoming #207). I'll have to think how to handle this. At first I thought we could just add a PREFIX class variable, but in case of the OpenAI model, there's already multiple. So I'm actually not sure if we can simply list all of these. So perhaps instead every model class should have a `list_models` method which we then combine.

klieret (changes requested):
.

hminsky2002:
@klieret

> One of the problems is that some models require a certain prefix that they strip (see for example the upcoming #207). I'll have to think how to handle this. At first I thought we could just add a PREFIX class variable, but in case of the OpenAI model, there's already multiple. So I'm actually not sure if we can simply list all of these. So perhaps instead every model class should have a `list_models` method which we then combine.

Ah, so for example 
'azure:gpt-4<etc>'  when used as the argument for OpenAIModel will produce an instance of the Azure client, as opposed
to 'gpt-4<etc>' which will just produce the OpenAI client, and we want to distinguish that both are available models?

hminsky2002:
> @klieret
> 
> > One of the problems is that some models require a certain prefix that they strip (see for example the upcoming #207). I'll have to think how to handle this. At first I thought we could just add a PREFIX class variable, but in case of the OpenAI model, there's already multiple. So I'm actually not sure if we can simply list all of these. So perhaps instead every model class should have a `list_models` method which we then combine.
> 
> Ah, so for example 'azure:gpt-4' when used as the argument for OpenAIModel will produce an instance of the Azure client, as opposed to 'gpt-4' which will just produce the OpenAI client, and we want to distinguish that both are available models despite the models dict only containing references to gpt-4 after the fact?

Maybe then the approach is to create an abstract 'list-models' function on the BaseModel class(as per your idea), that handles for any number of prefixes, then define them in each model instance with the relevant prefixes and append them to the given models? If that sounds like the right approach I'd love to take another crack at this!

hminsky2002:
@klieret I tried out the aforementioned approach and it seems fruitful, if you get a chance to look it over would love to know what you think!

ofirpress:
@klieret can we merge or close please?

klieret:
@ofirpress The whole model selection thing is a bit messy at the moment. Cannot merge this as-is (also because we since have diverged from it). Planning to get back to this after the next updates to the agent.

@hminsky2002 Thanks again for this! I'll probably take another look at this in the coming weeks and will try start from your commit so you get partial authorship on it, but there's also a few changes that we probably need to make.

hminsky2002:
> @ofirpress The whole model selection thing is a bit messy at the moment. Cannot merge this as-is (also because we since have diverged from it). Planning to get back to this after the next updates to the agent.
> 
> @hminsky2002 Thanks again for this! I'll probably take another look at this in the coming weeks and might try start from your commit so you get partial authorship on it, but there's also a few changes that we probably need to make.

No worries! Thanks for taking the time to look it over, happy to have tried it out! partial authorship would be cooooool but no sweat either way B)

PR #242, opened 2024-04-16
Fixed a potential error

DanjieTang:
I've ran into this error several times, where it says model_patch can't be None and ending the entire program.


klieret (approved):
Thanks! I've made it so that it doesn't even attempt to something instead

PR #258, opened 2024-04-18
Add hooks to agent

klieret:
Adds hooks to the `Agents` class that are going to be used for the web server. Prototype will be in #240


sweagent/agent/agents.py
```diff
@@ -747,14 +783,21 @@ def run(
...
         info = {}
         while not done:
+            for hook in self.hooks:
+                hook.on_step_start()
             state = env.communicate(self.state_command) if self.state_command else None
             thought, action, output = self.forward(
                 observation, env.get_available_actions(), state
             )
+            for hook in self.hooks:
+                hook.on_actions_gnerated(thought=thought, action=action, output=output)
```
klieret, line 800: fixme

PR #274, opened 2024-04-25
deprecate action from get_submission

khangich:
#### Reference Issues/PRs

- action is declared but never being used. Deprecate action argument from get_sumission

#### Any other comments?

🧡 Thanks for contributing!


klieret:
thx

PR #275, opened 2024-04-25
use shellcheck to check syntax

khangich:
#### Reference Issues/PRs

- avoid executing command when we check syntax. Executing command can increase overhead, CPU cycle and make the run time less efficient. Using shellcheck is likely sufficient and can save computer cycle. 

#### Any other comments?

🧡 Thanks for contributing!


klieret:
> Executing command can increase overhead, CPU cycle and make the run time less efficient. 

`bash -n` does not execute commands but only checks them for syntax errors.

It's not clear to me that shellcheck would save times. Moreover, shellcheck has a lot of code quality features that would be irrelevant and might confuse the model if we ask it to fix them. 

This is very unlikely to get merged without extensive studies.

PR #276, opened 2024-04-25
error handling get_pids

khangich:
#### Reference Issues/PRs

- add error handling when ps command fail
- filter out ps process  

#### Any other comments?

🧡 Thanks for contributing!


klieret:
Thank you, but without a clear problem that we're solving here, I'd rather leave the function as it currently is.

Btw, if you do the `try.. except .. as e ... raise AnotherException`, it's best practive to do `raise AnotherException from e` so that both exceptions show up properly :)

PR #277, opened 2024-04-25
refactor install env function

khangich:

- refactor install-env function 
- add util function to check if swe-bench repo

#### Any other comments?

🧡 Thanks for contributing!


klieret:
This seems to have been missing the main `_install_repo` method. While I agree that a refactoring is required (in particular after the recent changes), it requires very rigorous testing due to the many different ways that the environment can be specified, so we're going very slowly on this.

klieret:
Generally, I also wouldn't encourage refactorings that aren't motivated by added functionalities/bugfixes (because with imperfect test coverage, they can always introduce new problems)

PR #278, opened 2024-04-25
use git grep for faster search

khangich:

- Use git grep for better performance. Git grep leverages the Git index and can reduce the time required to search code in a GitHub repository, especially for large repositories.

#### Any other comments?

🧡 Thanks for contributing!


klieret:
Good idea. However, we'd have to actually test this to avoid subtle issues causing differences in performance (for example, `git grep` will not search any files that the agent has created itself).

So I think it's quite likely that this PR will hang for a while.

khangich:
Thanks. It makes sense.

PR #283, opened 2024-04-30
simplify close and make communicate method simpler

khangich:

- Enhance code readability by removing the requirement to call communicate(input="exit").
- Explicitly call terminate container instead of calling communicate. 
- The benefit is to make the communicate logic as lean as possible

#### Any other comments?

🧡 Thanks for contributing!


klieret:
I don't like this because `_terminate_container` is yet another method that sounds similar to `_close`, `_reset`, etc. Thank you for the effort, but I also wouldn't encourage refactorings that aren't motivated by added functionalities/bugfixes (because with imperfect test coverage, they can always introduce new problems)

PR #284, opened 2024-04-30
define exit actions

khangich:
…on in the future

- consolidate exit actions into one place. Make it easier to maintain and less error prone

#### Any other comments?

🧡 Thanks for contributing!


klieret:
Would only make it less error prone if they are used in that way in other parts.

PR #289, opened 2024-05-01
Add helpful message about conda env activation

klieret:


ofirpress:
Great to have stuff like this in the repo, thanks!

PR #291, opened 2024-05-01
Codespaces: Use pip install instead of creating new conda env

klieret:


Josephrp (approved):
pip instal --editable .  works great

PR #332, opened 2024-05-10
Fix typo in inspector server.py

nims11:
The typo leads to "Results format not recognized" error whenever viewing the eval report for a trajectory.


klieret:
Hmm, I don't think this is correct. I searched through the codebase and it's `not_generated` that gets appended, not `no_generation`. However, I have the feeling that the logical connection between the tests isn't correct (shouldn't it be `or` instead of `and`?)

klieret:
Could you open an issue where you explain exactly what you ran?

klieret:
Ah, looking at https://github.com/swe-bench/experiments/blob/main/evaluation/test/20240402_sweagent_gpt4/results/results.json, this might indeed be true. Perhaps this is from two different swe-agent versions...

klieret:
Thanks again, should be solved now!

Issue #328, opened 2024-05-09
Auto-scrolling in web-UI should only be active if user hasn't scrolled up.


klieret:
Every time a new message appears in one of the feeds/logs of the web UI, we scroll to the bottom. However, if a user wants to read a certain part during the run and has scrolled up, we shouldn't scroll down.

> [!tip] 
> General tips for frontend contributors:
> * Please open a separate PR for every issue that you address. Try to keep PRs small and concise (if you're thinking of bigger refactorings, please let us know)
> * In the web UI under "Extra Settings" there's a switch to run without needing API credits for testing purposes
> * Please install pre-commit (`pip install pre-commit && pre-commit install`) to ensure that your submission follows the styling guidelines

ofirpress:
+1

ofirpress:
@klieret we can close this, right?

klieret:
Thanks for the reminder, yes

ofirpress:
ty

PR #420, opened 2024-05-27
Issue #328: Auto-scrolling in web-UI Update

anishfish2:
Issue #328: 
- "Every time a new message appears in one of the feeds/logs of the web UI, we scroll to the bottom. However, if a user wants to read a certain part during the run and has scrolled up, we shouldn't scroll down."

Solution: 
- Implemented useRef's to track whether a user has scrolled in either EnvFeed, AgentFeed, or LogPanel. If a user has scrolled, at each update per feed/panel, do not auto-scroll. Otherwise, if a user has not scrolled or has scrolled to the bottom, at each update per feed/panel, continue to auto-scroll.
- Removed basic scrolldown feature from EnvFeed.js, AgentFeed.js, and Run.js for redundancy.

### DIFF
--- a/sweagent/frontend/src/Run.js
+++ b/sweagent/frontend/src/Run.js
@@ -208,7 +208,6 @@ function Run() {
           },
         ]);
         if (envFeedRef.current) {
-          console.log("checking env");
           setTimeout(() => {
             scrollEnv();
           }, 100);
@@ -225,7 +224,6 @@ function Run() {
         ]);
         if (agentFeedRef.current) {
           setTimeout(() => {
-            console.log("checking agent");
             scrollAgent();
           }, 100);
         }
--- a/sweagent/frontend/src/Run.js
+++ b/sweagent/frontend/src/Run.js
@@ -169,7 +169,8 @@ function Run() {
     }
   };
 
-  const scrollDetected = () => checkScrollPosition(logsRef, isLogScrolled, 58);
+  const scrollDetectedLog = () =>
+    checkScrollPosition(logsRef, isLogScrolled, 58);
   const scrollLog = () => scrollToBottom(logsRef, isLogScrolled);
 
   const scrollDetectedEnv = () =>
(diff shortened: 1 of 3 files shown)


sweagent/frontend/src/Run.js
```diff
@@ -154,8 +157,44 @@ function Run() {
...
+      ref.current.scrollHeight;
+  };
+
+  const scrollToBottom = (ref, scrollStateRef) => {
+    if (!scrollStateRef.current) {
+      ref.current.scrollTop = ref.current.scrollHeight;
+    }
+  };
+
+  const scrollDetected = () => checkScrollPosition(logsRef, isLogScrolled, 58);
```
klieret, line 172: why does the log need the offset and the others don't?

klieret: probably also should be called `scrollLogDetected` to make the naming uniform

sweagent/frontend/src/Run.js
```diff
@@ -154,8 +157,44 @@ function Run() {
...
+    checkScrollPosition(envFeedRef, isEnvScrolled);
+  const scrollEnv = () => scrollToBottom(envFeedRef, isEnvScrolled);
+
+  const scrollDetectedAgent = () =>
+    checkScrollPosition(agentFeedRef, isAgentScrolled);
+  const scrollAgent = () => scrollToBottom(agentFeedRef, isAgentScrolled);
+
   // Use effect to listen to socket updates
   React.useEffect(() => {
+    logsRef.current.removeEventListener("scroll", scrollDetected);
```
klieret, line 185: why are we first removing and then adding the same listener?

klieret: Actually, why do we need the listeners at all? Can't we just put this logic directly in the scroll functions?

klieret (approved):

sweagent/frontend/src/Run.js
anishfish2: The log needs the offset due to the copy to clipboard button that appears at the end. Updated the name to be scrollLogDetected

sweagent/frontend/src/Run.js
anishfish2: Removed the extraenous remote listeners commands. I believe we need the event listeners in order to continually update the state of whether the user has scrolled. Without it, I tried inserting the logic into the scroll functions, but had difficultly updating whether the user had scrolled or not depending on large message updates that would require explicitly saving past scroll positions. Please let me know if there is another solution or if you would recommend I continue to pursue this method.

CaptPreacher (approved):

sweagent/frontend/src/Run.js
klieret: Sounds good to me!

klieret (approved):

ofirpress:
Thanks @anishfish2 for making a PR. Thanks @klieret for reviewing & feedback!

Issue #478, opened 2024-06-01
Web UI: The `start_web_ui` script does not clean up its Flask processes on exit.


kwight:
### Describe the feature

The `start_web_ui.sh` script starts a Flask server via `sweagent/api/server.py`, but does not clean up those processes on exit (like it does with the React frontend app). This leaves port 8000 occupied, so running the script again will automatically fail with an "Address already in use" error, until those processes are killed manually (like with `lsof -i :8000`).

### Potential Solutions

Possible fixes:
1. Either `server.py` or the `start_web_ui` script could be modified to verify port 8000 is available, and use another untraditional port if needed.
2. Kill the Flask processes as part of `stop_react` (which could be renamed to a more generic `cleanup` or `shutdown`).

PR #479, opened 2024-06-01
Web UI: Kill the Flask server on exit.

kwight:
#### Reference Issues/PRs
Fixes https://github.com/princeton-nlp/SWE-agent/issues/478

This PR uses the PID returned when starting the Flask server to later clean up those processes on exit. This can avoid having to clean up manually before running the `start_web_ui` script again, or having to handle alternate ports or existing processes.

#### Testing Instructions
1. Run `./start_web_ui.sh`, then exit out of it.
2. Use `lsof -i :8000` to verify there are no processes on that port.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -2,15 +2,15 @@
 
 > How can I change the demonstrations given to SWE-agent?
 
-At the start of each run, we feed the agent a demonstration trajectory, showing it how to solve an example issue. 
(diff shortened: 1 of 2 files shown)


klieret:
Thanks for fixing this! (previously the flask command was run in the foreground and I didn't think of changing the cleanup!)

Issue #485, opened 2024-06-02
Web UI: Markdown is not being rendered.


kwight:
### Describe the bug

We have `react-markdown` installed but don't seem to be using it anywhere, and we can see that backticks are rendered literally in the agent feed.

<img width="1235" alt="Screenshot 2024-06-01 at 5 03 48 PM" src="https://github.com/princeton-nlp/SWE-agent/assets/349751/7419f904-d744-4f95-9d90-af7654f2f76e">

### Steps/commands/code to Reproduce

Run a job in the web UI and wait for agent feed content that includes backticks (`).

### Error message/results

I would expect content within backticks to be styled, particularly since the `react-markdown` package is present.

### System Information

Mac Studio, M2 Max chipset, 96GB RAM
MacOS Sonoma 14.5
Python 3.12.2

### Checklist

- [X] I'm running with the latest docker container/on the latest development version
- [X] I've searched the other issues for a duplicate
- [X] I have copied the full command/code that I ran (as text, not as screenshot!)
- [X] If applicable: I have copied the **full** log file/error message that was the result (as text, not as screenshot!)
- [X] I have enclosed code/log messages in triple backticks ([docs](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#quoting-code)) and clicked "Preview" to make sure it's displayed correctly.

PR #486, opened 2024-06-02
Web UI: Render Markdown in agent feed messages.

kwight:
#### Reference Issues/PRs
Fixes https://github.com/princeton-nlp/SWE-agent/issues/485

This PR uses the already-installed `react-markdown` package to render agent feed messages as markdown.

| Before | After |
| --- | --- |
| Screenshot 2024-06-01 at 5 03 48 PM | Screenshot 2024-06-01 at 5 21 02 PM |

#### Testing Instructions
1. Run `./start_web_ui.sh`.
2. Set a problem source and click Run.
3. As content fills the agent feed (the "Thoughts" section), verify it's rendered as markdown, and not literally.

### DIFF
--- a/sweagent/frontend/src/components/AgentMessage.js
+++ b/sweagent/frontend/src/components/AgentMessage.js
@@ -3,6 +3,7 @@ import React from "react";
 import "../static/message.css";
 import "../static/agentMessage.css";
 import { Gear } from "react-bootstrap-icons";
+import Markdown from "react-markdown";
 
 const AgentMessage = ({
   item,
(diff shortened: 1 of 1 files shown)


ofirpress:
cool idea, thanks

klieret:
> cool idea, thanks. we could also add markdown formatting to the demonstration if it doesn't have any right now, that would make this even better

The demonstrations are currently not shown in the agent feed (but we could add them at some point).

klieret:
Very nice @kwight , that's a great addition ❤️

ofirpress:
> The demonstrations are currently not shown in the agent feed (but we could add them at some point).
No @klieret I dont want to show the demo in the web UI. 

what i wanna do is modify the demo so that the 'Thoughts' use some markdown syntax here and there, that way the agent will know that it's allowed to do that and will maybe do it more often? meh, not super important

ofirpress:
actually nvm I see it's already kinda there.

PR #592, opened 2024-06-19
Fixed #571 - show message that model arg is ignored in case of using Azure OpenAI

### LINKED ISSUE #571: When using the azure GPT endpoint, the --model CLI arg is ignored
hi!
When using the azure GPT endpoint, the --model CLI arg is ignored

So we should give a warning (or an error?) if the user tries using it.

(we instead get the model from: OPENAI_API_VERSION: '...')
see https://github.com/princeton-nlp/SWE-agent/blob/main/sweagent/agent/models.py#L247

jank:
…AzureOpenAI

#### Reference Issues/PRs
Fixes #571 

Log warning message that model command line arg is ignored in case of using AzureOpenAI.

### DIFF
--- a/sweagent/agent/models.py
+++ b/sweagent/agent/models.py
@@ -244,6 +244,11 @@ class OpenAIModel(BaseModel):
 
         # Set OpenAI key
         if self.args.model_name.startswith("azure"):
+            if 'model' in self.args and self.args.model:
+                logger.warning(
+                    "The --model CLI argument is ignored when using the Azure GPT endpoint. "
+                    "The model is determined by AzureOpenAI.",
+                )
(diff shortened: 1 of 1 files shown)


klieret (approved):
Thanks ❤️ ! (still planning to avoiding this problem in general, but this is an improvement for the here and now)

jank:
Just for the record. I created the original diff for this fix using SWE-Agent.
A few manual changes were necessary as GPT-4 introduced an import of logging instead of using the available logger. 
Overall quite impressive work from the SWE-Agent team.

ofirpress:
Cool! Thank you so much!

PR #619, opened 2024-06-27
Add filemap command in the spirit of repomap

samuela:
#### Reference Issues/PRs
n/a

Adds a "filemap" command to aid in exploring large files/codebase and avoid `scroll_down`-hell. For example,
```
$ filemap SWE-agent/sweagent/agent/agents.py
...
   239 class Agent:
   240     """Agent handles the behaviour of the model and how it interacts with the environment."""
   241 
   242     def __init__(self, name: str, args: AgentArguments):
   243 ... eliding lines 243-256 ...
   257 
   258     def add_hook(self, hook: AgentHook):
   259         """Add hook to agent"""
   260         hook.on_init()
   261         self.hooks.append(hook)
   262 
   263     def _append_history(self, item: dict):
   264         for hook in self.hooks:
   265             hook.on_query_message_added(**item)
   266         self.history.append(item)
   267 
   268     def setup(self, instance_args, init_model_stats=None) -> None:
   269 ... eliding lines 269-322 ...
...
```
It is currently designed to elide any function or method definition that is at least 5 lines long. Changing this setting is straightforward.

There is currently no special handling for docstrings. If a function has a docstring it will be elided by default (unless the entire function body is less than 5 lines long).

Remaining TODOs:
- [ ] I have not verified that this command works within the SWE-agent docker environment, and I am not yet sure what the right approach is to test this.

### DIFF
--- a/config/commands/filemap.py
+++ b/config/commands/filemap.py
@@ -1,0 +1,54 @@
+# @yaml
+# signature: filemap <file_path>
+# docstring: Print the contents of a Python file, skipping lengthy function and method definitions.
+# arguments:
+#   file_path:
+#       type: file path
+#       description: The path to the file to be read
+#       required: true
+
+from __future__ import annotations
+
+import argparse
+import warnings
+
+# tree_sitter is throwing a FutureWarning
+warnings.simplefilter("ignore", category=FutureWarning)
+from tree_sitter_languages import get_language, get_parser
+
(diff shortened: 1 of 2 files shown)


klieret:
Awesome :D

The easiest way to test this is to just do `py run.py ... --model human` and then you can type commands instead of the AI. So you could try what filemap does with some of the issues

PR #621, opened 2024-06-27
Cap swebench version dependency

carlosejimenez:
Caps the `swebench` package dependency `<2.0.0` to avoid breaking changes in the major version update.

We should update SWE-agent to use the new `swebench` package in the near future, but this should avoid issues with new users of SWE-agent.


klieret (approved):
Thanks!

PR #625, opened 2024-07-01
recommended way to upgrade pip

hubstrauss:
- safe method for pip upgrade
- adding an example on how to pass azure model


klieret (approved):

klieret:
Thanks @hubstrauss

PR #628, opened 2024-07-02
ensure variables work in special command docstring

forresty:
SYSTEM message changes:

before: moves the window down {WINDOW} lines
after: moves the window down 100 lines

I was playing it and found that in special commands section in the system message, `{WINDOW}` was not replaced with the desired value of `100`. Dig into it a little bit and I believe this should fix it. Not sure if there would be side effects tho.

image

### DIFF
--- a/sweagent/agent/commands.py
+++ b/sweagent/agent/commands.py
@@ -243,7 +243,7 @@ class ParseCommandDetailed(ParseCommandBash):
         for cmd in commands + subroutine_types:
             docs += f"{cmd.name}:\n"
             if cmd.docstring is not None:
-                docs += f"  docstring: {cmd.docstring}\n"
+                docs += f"  docstring: {cmd.docstring.format(**kwargs)}\n"
             if cmd.signature is not None:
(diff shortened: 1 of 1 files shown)


klieret (approved):
This looks good to me. In fact what you propose was almost certainly intended (we see it being used in the `generate_command_docs` of the super class.

This is how it's being called:

```python
        object.__setattr__(
            self,
            "command_docs",
            self.parse_command.generate_command_docs(
                self._commands,
                self.subroutine_types,
                **self.env_variables,
            ),
        )
```

which is in the `__post_init__` of the `AgentConfig`, so the `env_vars` there are what's being defined in the config, e.g.,

```yaml
env_variables:
  WINDOW: 100
  OVERLAP: 2
  CURRENT_LINE: 0
  CURRENT_FILE: ''
  SEARCH_RESULTS: ()
  SEARCH_FILES: ()
  SEARCH_INDEX: 0
```

klieret:
Very nice contribution ❤️

forresty:
that was fast! thanks!

PR #668, opened 2024-07-16
Add linter for JS, HTML, bash

### LINKED ISSUE #664: Add linters for other programming languages
Currently we only have linting support for python.

https://github.com/princeton-nlp/SWE-agent/blob/main/config/commands/edit_linting.sh#L74

We've been asked for this a few times: we should add linting support for other languages. Based on the StackOverflow survey, these are the 10 languages that are most popular other than Python:

Javascript, HTML/CSS, SQL, TypeScript, Bash, Java, C#, C++, C, PHP

We should add support for linting for this. We're going to have to make individual considerations for every language. 

That's it.

(For reference, see this closed bug report that's a bit less specified than this one: https://github.com/princeton-nlp/SWE-agent/issues/53 )

zachblume:
#### Reference Issues/PRs

Fixes #664 

Expands inferred lint command from flake8 to flake8/eslint/htmlhint/shellcheck depending on file extension

### DIFF
--- a/config/commands/edit_linting.sh
+++ b/config/commands/edit_linting.sh
@@ -46,7 +46,30 @@ edit() {
         return
     fi
 
-    local linter_cmd="flake8 --isolated --select=F821,F822,F831,E111,E112,E113,E999,E902"
+    # Prep a JSX-enabled config file for eslint 9
+    cat <<EOF > eslint.config.mjs
+// eslint.config.js
+export default [
+    {
+        "languageOptions": {
+            "parserOptions": {
+                "ecmaFeatures": {
+                    "jsx": true
+                }
+            }
+        }
+    }
+];
+EOF
+    
+    # Infer linter based on file extension
+    case "$CURRENT_FILE" in
+        *.py) linter_cmd="flake8 --isolated --select=F821,F822,F831,E111,E112,E113,E999,E902" ;;
+        *.[jt]sx?) linter_cmd="eslint ."
+        *.html?) linter_cmd="htmlhint --format unix" ;;
+        *.sh) linter_cmd="shellcheck" ;;
+        *) linter_cmd="" ;;
+    esac
     local linter_before_edit=$($linter_cmd "$CURRENT_FILE" 2>&1)
 
     # Bash array starts at 0, so let's adjust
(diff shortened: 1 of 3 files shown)


klieret:
Hi @zachblume , thanks for getting this started. 

A general note though: We shouldn't blindly add linters, but only select flags that indicate that the code is really wrong. For example indentation issues in python. 

`shellcheck` and the like will include tons of optional checks that are more code quality/style improvements rather than things that will necessarily fail the solution.

So we should go slow on the linters and only enable the most basic checks.

zachblume:
Waking I know best and will include far too much if you pull in js/recommended but is pretty calm by default. I found you can no longer reliably run the newest version of eslint without a config file when attempting to apply rules and because JSX is off by default, pulling in a config with JSX:enabled is necessary if you want any React codebase to not fail lint.

Htmlhint and shell check I know less and so will modify to specify rules if possible; I saw that is how you all approached flake8 with a rule list as well

sweagent/environment/swe_env.py
```diff
@@ -373,8 +373,10 @@ def reset(self, index: int | None = None, apply_test_patch: bool = False) -> tup
         # Call install environment helper function if specified
         if self.install_environment:
             self.install_env()
-        # Install mypy for linting purposes
+        # Install linters
         self.communicate_with_handling("pip install flake8", error_msg="Failed to install flake8 (lint library)")
+        self.communicate_with_handling("npm install -g eslint", error_msg="Failed to install eslint (lint library)")
```
klieret, line 378: I think the reason we install flake8 directly after the container is initialized is because we switch to a conda env (I think it might actually be better to install it in a global python env and then specify the abspath to there, but not important).

But with eslint etc., there's no reason to not directly install it to the docker, right?

These commands will run for every new instance, so we really want to have as few as possible here.

config/commands/edit_linting.sh
```diff
@@ -46,7 +46,30 @@ edit() {
         return
     fi
 
-    local linter_cmd="flake8 --isolated --select=F821,F822,F831,E111,E112,E113,E999,E902"
+    # Prep a JSX-enabled config file for eslint 9
+    cat <<EOF > eslint.config.mjs
```
klieret, line 50: won't that pollute the git dir that we're currently in? Probably should put the config somewhere else and then specify the path to it...

klieret (changes requested):

klieret:
Thanks again for this, but let me close this. SWE-agent M will generalize to JS and also has some form of linting (but it's always a delicate balance with these linters because you don't want to block too much)

PR #698, opened 2024-07-23
Update coding_challenges.md - add git init command

ofirpress:


ofirpress:
@klieret this is correct, right?

klieret:
Oh true! Though we also need to have a first commit in it, I think. Lm do that

ofirpress:
This is based on a conversation with @msaroufim , all credit to Mark :) 

Thanks @klieret

Issue #720, opened 2024-08-15
Groq Models


MohammedNagdy:
### Describe the feature

Add new models (open source) that utilizes Groq fast TPU inference methods. Hosted on Groq platform.

### Potential Solutions

Add the Groq models based on the OpenAI model in the `sweagent/agent/models.py`. Check [Groq docs](https://console.groq.com/docs/quickstart).

PR #721, opened 2024-08-15
Groq Models Integration

MohammedNagdy:
**Groq Models Integration**

Benefits:
- Open source models with increased inference speed
- Free for low usage espacially small projects

Usage example
```bash
$ python run.py --model_name groq/llamainstant8  \
    --data_path https://github.com/princeton-nlp/SWE-agent/issues/666  \
    --repo_path ./ \
    --config_file config/default_from_url.yaml  \
    --apply_patch_locally
```

This PR closes [Groq Models Issue](https://github.com/princeton-nlp/SWE-agent/issues/720)


klieret (changes requested):
I'd really just rather add `litellm` support as pointed out at previous PRs... But still, since I didn't have time to do that yet, I'm open to just merge that once you fix the test (just add the `groq` dependency).

Still, don't expect good performance from such small open source models, at least for now.

ofirpress:
I don't think we should merge support for small models for now. Willing to be convinced otherwise though, but I would wanna see results on SBL

MohammedNagdy:
@ofirpress Thanks for your comment. However, I would like to point out that `groq` models are not just basic small open source models groq will provide insane speed to the agent as it will increase inference speed. I am also communicating to them on adding more new opensource models like [codestral mamba](https://mistral.ai/news/codestral-mamba/). Let me know your opinion too.

MohammedNagdy:
@ofirpress Thanks for your comment. However, I would like to point out that `groq` models are not just basic small open source models groq will provide insane speed to the agent as it will increase inference speed. I am also communicating to them on adding more new opensource models like [codestral mamba](https://mistral.ai/news/codestral-mamba/). Let me know your opinion too.

klieret (approved):
I'm just gonna merge this for now, because it's tested and doesn't make things more complicated. But we really should just add support for litellm and then remove all this. 

@MohammedNagdy do you want to add litellm support? Should be relatively easy and then we'd just support everything. See #64

MohammedNagdy:
@klieret I will take a look!

ofirpress:
I'm not saying we should remove this, but I'm kind of opposed to continuing to add more support for small models, because people will try using swe-agent with these small models, it won't work, and they'll be annoyed and will not try bigger models.

ofirpress:
anyways having litellm is the right solution, but also when we do finally put in litellm we should maybe issue a warning when a small model is used.

PR #739, opened 2024-08-22
Add GitLab support + AWS Bedrock Claude-3.5 and keys to keys.cfg

jpaodev:
This patch includes changes to add GitLab support to SWE Agent. (E.g. for companies that use GitLab SaaS)

#### Reference Issues/PRs
N/A

- Add support for GitLab issue URLs and repository URLs. GitLab endpoint can be configured using `GITLAB_URL` variable in `keys.cfg`. Gitlab Token can be configured using `GITLAB_TOKEN` in `keys.cfg`.
- Added support to create GitLab pull requests once patch is available and `--open_pr` has been used.
- Adding GitLab support introduced new dependency `python-gitlab` to `requirements.txt`
- Added Claude 3.5 Sonnet to available AWS Bedrock models
- Added possibility to supply `AWS_ACCESS_KEY`, `AWS_SECRET_KEY` and `AWS_REGION` to `keys.cfg` if desired.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -20,3 +20,4 @@ flask
 flask-cors
 flask-socketio
 groq
+python-gitlab
--- a/run.py
+++ b/run.py
@@ -46,10 +46,14 @@ from sweagent.agent.models import ModelArguments
 from sweagent.environment.swe_env import EnvironmentArguments, SWEEnv
 from sweagent.environment.utils import (
     InvalidGithubURL,
+    InvalidGitlabURL,
     get_associated_commit_urls,
+    get_associated_commit_urls_gitlab,
     get_data_path_name,
     get_gh_issue_data,
+    get_gl_issue_data,
     parse_gh_issue_url,
+    parse_gl_issue_url,
 )
 
 __doc__: str = """ Run inference. Usage examples:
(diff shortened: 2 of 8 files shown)


jpaodev:
Example command to run with a GitLab repository:
```
python run.py \     
  --model_name azure:gpt-4o-mini \
  --data_path https://gitlab.com/jpaodev/test-project-fix/-/issues/3 \
  --repo_path https://gitlab.com/jpaodev/test-project-fix \
  --config_file config/default_from_url.yaml \
  --per_instance_cost_limit 2.00 --open_pr
 ```

This also works with custom GitLab deployment URLs.

jpaodev:
Accidental close, added commit to pull request (removed hardcoded assertion)

klieret (changes requested):
Hi @jpaodev . Thanks a lot for all the work that went into this ❤️ . It's a bit difficult to review PRs that are that large and that mix different objectives, so let me split it up here:

1. Adding AWS bedrock keys: Sure, sounds useful. Would merge this immediately.
2. Adding Sonnet for AWS bedrock: Sure (but I really want to get away from all the hardcoded models and just switch to using #64 -- this would be a really useful thing to work on). Would merge this immediately
3. Adding `open_pr` to the web UI: Very nice :)  Would merge immediately.
4. I'm less sure about the gitlab support, because we were trying to keep this as simple as possible (and github is just much more popular for most OSS). However, since you already worked on this to this extend, I'm open to consider it. However, this would at least need unittests that cover all the gitlab lines. 

Could you perhaps open separate PRs for these 4 things? Also helps with bookkeeping in the changelog etc.

If it's too hard to separate 1.-3., creating one PR with 1.-3. would also be ok (but luckily it's mostly touching different files or you could alternatively work with `git add -p`). But let's look at 4. separately because it probably needs a bit more work.

Again, thanks a lot, much appreciated (I'd just recommend to open an issue for discussion before investing a lot of work in a PR in the future -- just to avoid situations where maintainers have to reject)

klieret:
Let me close this based on my earlier comment

PR #751, opened 2024-08-29
CI: Use GITHUB_TOKEN from keys.cfg

### LINKED ISSUE #737: Tests do not use github token from `keys.cfg`
only read it as an env variable

klieret:
Closes #737


Monalmos:
https://github.com/princeton-nlp/SWE-agent/pull/751/commits/300215a295dd270afc948091fd31ffff5e24edff

PR #792, opened 2024-10-04
Update README.md

kushal34712:
fixed style for the YouTube video 
added contributors list as their profile icon 
fixed some punctuation


klieret:
thanks, but those are some semi-arbitrary changes.

PR #793, opened 2024-10-04
Refactor file handling to use pathlib

### LINKED ISSUE #445: Replace `os.path` with `pathlib`
`os.path` is cumbersome. Let's default to using `pathlib` everywhere. Most conversions can probably be done by copilot.

Once done, add [these ruff checks](https://docs.astral.sh/ruff/rules/#flake8-use-pathlib-pth) to make sure it stays that way.

sravan1946:
#### Reference Issues/PRs

Fixes #445 

I have added the ruff checks and made changes so that all checks pass.

### DIFF
--- a/inspector/server.py
+++ b/inspector/server.py
@@ -282,7 +282,6 @@ def main(data_path, directory, port):
             with Path.open(Path(data_path)) as f:
                 data = json.load(f)
     elif "args.yaml" in os.listdir(directory):
-        # with Path.open(os.path.join(directory, "args.yaml")) as file:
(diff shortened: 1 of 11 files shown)


klieret:
Hi, thank you for your work :)  However, this isn't quite the way to use `pathlib` concisely. Rather, the objects that are passed around should be made the `Path` type `path = Path("/some/path/")` and then you can directly call all the operations on them `path.parent.mkdir()` etc. Just replacing built-in open with `Path.open` makes it more cluttered.

PR #797, opened 2024-10-05
Chore: Replace os.path with pathlib and update ruff lint rules

### LINKED ISSUE #445: Replace `os.path` with `pathlib`

Prathamesh010:
#### Reference Issues/PRs

Fixes #445

Removed all instances of os.path and used pathlib


ofirpress:
Thanks for contributing!

PR #824, opened 2024-10-23
Improved Clarity in Demonstration, Summarizer, and Command Configuration

Mefisto04:
Corrected minor phrasing issues in demonstration, summarizer, and command configuration docs for clarity.
- The phrase "Observation output exceeding these number, will be summarized" corrected to "Observation output exceeding this number will be summarized."
- "we only ingest demonstrations in the from of a trajectory file" corrected to "we only ingest demonstrations in the form of a trajectory file."
- "You can also leverage third party libraries (check out how we do linting enabled edit in config/commands/edit_linting.sh)" corrected to "You can also leverage third-party libraries (check out how we enable linting in config/commands/edit_linting.sh)."


klieret:
Thx

PR #843, opened 2024-11-27
Run SWE-agent on Commit0

nanjiangwill:
#### Reference Issues/PRs
[Commit0](https://github.com/commit-0/commit0/tree/main) is a from scratch AI coding challenge.

The benchmark consists of 57 core Python libraries. The challenge is to rebuild these libraries and pass their unit tests. All libraries have:

This PR adds necessary functions to run commit0 with SWE-agent but run into bugs

#### Issues
1. Without using any tool or command line and no hint in prompt/context, in response SWE-agent will just do "calling cmd and self-answered with wrong output", example in `commit0_prepare/traj_example/97af1e.traj`
2. sometimes, it will even generate(instead of actual running pytest) the test result

#### To Reproduce

1. run `python commit0_prepare/generate_prompt_for_each_repo.py`
2. run `commit0_prepare/run_commit0.sh` with API key

#### What is changed in source code
1. we use existing docker and have SWE-agent run on it, so we do not need to have `--repo_path` argument since files are already in docker
2. we are using `uv` in docker, so the env setup code is changed
3. commit0 docker need to run with platform flag specified


klieret:
(thank you! but currently closing old PRs and we never quite ended up going in this direction)

PR #857, opened 2024-12-10
Fixed  the spell, that may mislead model

acheshkov:
The default action "scroll_up" has inconsistent description, the same as for "scroll_down"

The proposed change fixes the spell in the description of the action


klieret (approved):
Oooh, very good catch, thank you!

PR #866, opened 2024-12-20
Add verbosity control via CLI flag

### LINKED ISSUE #501: Set verbosity of all logs with CL flag
The `--verbose` flag isn't great, let's just set `logging.setLevel` numbers. Also, since it's not distinguishable from the CLI, let's make `verbose` affect all logs, rather than just the env.

Should we have a debug level where you can print all commands issued in the env and their output?

* [ ] Should the log level affect the output files?

erkinalp:
This PR adds a --verbosity flag to control logging levels globally, addressing issue #501.

Key changes:
- Add --verbosity CLI flag with standard log level choices (DEBUG, INFO, WARNING, ERROR, CRITICAL)
- Add set_default_levels() function for dynamic level control
- Add unit tests for verbosity control
- Maintain backward compatibility with environment variables

The implementation allows users to control logging verbosity through:
1. CLI flag: --verbosity [DEBUG|INFO|WARNING|ERROR|CRITICAL]
2. Environment variables (existing): SWE_AGENT_LOG_STREAM_LEVEL, SWE_AGENT_LOG_FILE_LEVEL

Example usage:
```bash
# Set verbosity via CLI
sweagent run --verbosity DEBUG ...

# Set verbosity via environment (existing method)
export SWE_AGENT_LOG_STREAM_LEVEL=DEBUG
sweagent run ...
```

Link to Devin run: https://app.devin.ai/sessions/e14f693390ed47ba87d12698d5834d08

Fixes #501

### DIFF
--- a/sweagent/run/common.py
+++ b/sweagent/run/common.py
@@ -217,6 +217,7 @@ class BasicCLI:
 
     def get_config(self, args: list[str] | None = None) -> BaseSettings:
         """Get the configuration object from defaults and command arguments."""
+        from sweagent.utils.log import set_default_levels  # Import here to avoid circular imports
 
         # >>> Step 1: Use argparse to add a --config option to load whole config files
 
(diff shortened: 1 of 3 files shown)


klieret (changes requested):
Don't think this works because `--verbosity` is actually never used.

PR #871, opened 2024-12-21
feat: Add warning for ignored model argument with Azure endpoints

### LINKED ISSUE #571: When using the azure GPT endpoint, the --model CLI arg is ignored

erkinalp:
When using Azure endpoints, the --model CLI argument is ignored as the model is determined by OPENAI_API_VERSION. This change adds a warning message to inform users about this behavior.

Changes:
- Added warning in LiteLLMModel.__init__ when Azure endpoint is detected
- Created tests to verify warning behavior for Azure and non-Azure endpoints
- Ensures users are informed when their model argument will be ignored

Link to Devin run: https://app.devin.ai/sessions/57d1cf33292440e6ba72a487ab4e049b

Fixes #571

### DIFF
--- a/sweagent/agent/models.py
+++ b/sweagent/agent/models.py
@@ -431,6 +431,17 @@ class LiteLLMModel(AbstractModel):
         self.args = args
         self.stats = InstanceStats()
         self.tools = tools
+
+        # Check for Azure endpoint and warn about ignored model argument
+        if self.args.api_base and "azure" in self.args.api_base.lower():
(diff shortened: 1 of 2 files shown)


klieret:
#571 was already fixed when we switches to litellm

PR #883, opened 2025-01-23
Saves all preds under output_dir / instance.problem_statement.id dire…

carlosejimenez:
…ctories for run-batch


carlosejimenez:
Fixes #884

klieret (approved):

PR #966, opened 2025-02-19
Run Batch: Allow passing custom instances source config

RNabel:
Allow passing arbitrary implementations of `AbstractInstanceSource` by adding it to the union type in the pydantic model. It currently raises when trying to construct the RunBatchConfig with a different instances source.

#### Reference Issues/PRs
None

N/A

### DIFF
--- a/sweagent/run/batch_instances.py
+++ b/sweagent/run/batch_instances.py
@@ -308,4 +308,6 @@ class ExpertInstancesFromFile(BaseModel, AbstractInstanceSource):
         return self.path.stem
 
 
-BatchInstanceSourceConfig = InstancesFromHuggingFace | InstancesFromFile | SWEBenchInstances | ExpertInstancesFromFile
+BatchInstanceSourceConfig = (
+    InstancesFromHuggingFace | InstancesFromFile | SWEBenchInstances | ExpertInstancesFromFile | AbstractInstanceSource
+)


klieret:
I guess this sounds reasonable. But have you tested this? These union type things (especially with pydantic automatically picking the right one) have been a bit finnicky. 

> It currently raises when trying to construct the RunBatchConfig with a different instances source.

Can you explain a bit more what you were trying? I guess you were constructing a different pydantic model and then passing that via the python interface? (but then what was raising an error exactly? Or was it just that mypy or something flagged it?)

klieret:
> Can you explain a bit more what you were trying? I guess you were constructing a different pydantic model and then passing that via the python interface? (but then what was raising an error exactly? Or was it just that mypy or something flagged it?)

Because the 'config yaml file' to custom class might not be possible without adding the explicit types to the union (because how should pydantic know which model to initialize)

RNabel:
> Can you explain a bit more what you were trying?

Yep! I have a list of (env, prompt) in Python and would like to run them as a batch. From what I can tell, there isn't a `RunBatchConfig` which allows you to just provide a list of `BatchInstance`s - instead the existing ones all read from external data sources (files, etc.).

The solution was to add a super basic type that implements the abstract class, e.g.:
```
class Instances(BaseModel, AbstractInstanceSource):
    instances: list[BatchInstance]

    def get_instance_configs(self) -> list[BatchInstance]:
        return self.instances
```

When you pass this to the `RunBatchConfig`, it raises because the type isn't part of the union. So roughly:
```
RunBatchConfig(
        instances=Instances(instances=[...]),
...
)
```

Results in:
```
pydantic_core._pydantic_core.ValidationError: 4 validation errors for RunBatchConfig
instances.InstancesFromHuggingFace
  Input should be a valid dictionary or instance of InstancesFromHuggingFace [type=model_type, input_value=Instances(instances=[]), input_type=Instances]
    For further information visit https://errors.pydantic.dev/2.10/v/model_type
instances.InstancesFromFile
  Input should be a valid dictionary or instance of InstancesFromFile [type=model_type, input_value=Instances(instances=[]), input_type=Instances]
    For further information visit https://errors.pydantic.dev/2.10/v/model_type
instances.SWEBenchInstances
  Input should be a valid dictionary or instance of SWEBenchInstances [type=model_type, input_value=Instances(instances=[]), input_type=Instances]
    For further information visit https://errors.pydantic.dev/2.10/v/model_type
instances.ExpertInstancesFromFile
  Input should be a valid dictionary or instance of ExpertInstancesFromFile [type=model_type, input_value=Instances(instances=[]), input_type=Instances]
    For further information visit https://errors.pydantic.dev/2.10/v/model_type
```

Having written all of this, maybe it's worth adding a `SimpleInstances` class to make this workflow easier?

> how should pydantic know which model to initialize

yeah, this only fixes construction of these types from within Python, and doesn't help with loading these from files.

klieret:
Appreciate the clarification, that makes total sense!

> When you pass this to the RunBatchConfig, it raises because the type isn't part of the union. So roughly:

Oh true, because that's again validated by pydantic when we assign it to the other model, right!

> Having written all of this, maybe it's worth adding a SimpleInstances class to make this workflow easier?

But if you're using everything via python anyway, can't you just use the `RunBatch` constructor instead of `RunBatch.from_config`?   (basically I tried to both a full python interface where you pass around objects and then the `from_config` which initializes from serialized things)

Like can't you just do this:

```python
RunBatch(
    instances=[instance1, instance2, ...],
    agent_config=agent_config,
    ...
)
```

RNabel:
Ha, looks like I was holding this API wrong - thanks for the explanation!

PR #1009, opened 2025-03-05
tools, fix exit_forfeit signature

### LINKED ISSUE #1011: `exit_forfeit` not recognized
The `exit_forfeit` tool currently doesn't work. It was previously directly caught by `agents.py`, but we no longer do this, so it currently has no effect.

RNabel:
Summary: The signature is `exit_forfeit` but the code checking for it look for `exit`. i.e. see sweagent/agent/agents.py:895

This PR changes the signature to match (instead of changing the relevant checking logic).

Closes #1011

### DIFF
--- a/tools/forfeit/config.yaml
+++ b/tools/forfeit/config.yaml
@@ -1,5 +1,5 @@
 tools:
   exit_forfeit:
-    signature: "exit_forfeit"
+    signature: "exit"
     docstring: "Give up on the current challenge and terminate the session."
     arguments: []


klieret:
So the side effect of this will be that it will kill the shell session. For a long time, this would've been bad, because then we can't do the `autosubmit_after_error`, but I acutally think this should work, meanwhile. Still, the exit status would not be set to `exit_forfeit`, I think. So I'll probably instead go with `exit_forfeit` echoing a sentinal value and `agents.py` catching that...

klieret:
Closing in favor of #1024 

Thank you again @RNabel ! As mentioned going with a different way of catching this to ensure that we have a distinct exit status, but I really appreciate the issue & PR!

PR #1016, opened 2025-03-08
chore: Enable anthropic header for output-128k

LuisValdesZero:
I was using the model sonnet and I got into problems because of the output tokens being limited to 60000. I added this value output-128k-2025-02-19 to the anthropic-beta header.

### DIFF
--- a/sweagent/agent/models.py
+++ b/sweagent/agent/models.py
@@ -645,6 +645,9 @@ class LiteLLMModel(AbstractModel):
         completion_kwargs = self.config.completion_kwargs
         if self.lm_provider == "anthropic":
             completion_kwargs["max_tokens"] = self.model_max_output_tokens
+            completion_kwargs["extra_headers"] = {
+                'anthropic-beta': 'output-128k-2025-02-19'
+            }
         try:
             response: litellm.types.utils.ModelResponse = litellm.completion(  # type: ignore
                 model=self.config.name,


klieret:
Hi @LuisValdesZero thanks for this. I'm wondering, could you just use [`completion_kwargs`](https://swe-agent.com/latest/reference/model_config/#sweagent.agent.models.GenericAPIModelConfig.completion_kwargs) for this?

I'd happily merge a documentation PR though

klieret:
(I'm a bit curious that you actually need that, since this is about output tokens -- usually those are quite limited in our use of the agent)

More generally, should probably keep to the principle of least surprise for users, so I don't want to make something like this the default.

klieret:
Or was this because you were using 3.7 in high reasonsing mode and you get a lot of extended thinking output tokens?

LuisValdesZero:
Hello @klieret , I was running the model claude-3-7-sonnet-20250219 and I got some errors related to max token, then I saw some docs from [Anthropic](https://docs.anthropic.com/en/docs/about-claude/models/extended-thinking-models#extended-output-capabilities-beta) and [AWS](https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-anthropic-claude-37.html).

After some investigation I saw that it was not implemented and I created the patch.



> Extended output length (beta)
> 
> Claude 3.7 Sonnet can produce substantially longer responses than previous Claude models, with support for up to 128K output tokens (beta). This extended output length can be used with the new reasoning capabilities. This feature can be enabled by passing an anthropic-beta inference parameter of output-128k-2025-02-19.



> (I'm a bit curious that you actually need that, since this is about output tokens -- usually those are quite limited in our use of the agent, right? Whereas the context window/input tokens can be a limiting factor depending on how the trajectory goes)
> 
> More generally, should probably keep to the principle of least surprise for users, so I don't want to make something like this the default.

LuisValdesZero:
> Hi @LuisValdesZero thanks for this. I'm wondering, could you just use [`completion_kwargs`](https://swe-agent.com/latest/reference/model_config/#sweagent.agent.models.GenericAPIModelConfig.completion_kwargs) for this?

The extra_headers is an optional parameter for [litellm.completion ](https://docs.litellm.ai/docs/completion/input) and you can use it to send the additional headers. **completion_kwargs expands and the extra_headers is available for the function to use. I don't know how to explain in an easier way but it worked.

extra_headers: dict (optional) - Alternative to headers, used to send extra headers in LLM API request.

LuisValdesZero:
> More generally, should probably keep to the principle of least surprise for users, so I don't want to make something like this the default.

Okay, I guess leaving something like this as default might not be a good idea, I was just playing around and saw this might be useful, maybe better as a flag or something like that.

Issue #1022, opened 2025-03-12
run batch processing with modal


vsee:
### Describe the feature

In its current state it does not seem possible to run batch processing with modal.

```bash
sweagent run-batch \
    --config mymodelconfig.yaml \
    --agent.model.name mymodel/modelname \
    --instances.deployment.type=modal \
    --instances.type swe_bench \
    --instances.subset lite \
    --instances.split dev  \
    --instances.slice :3
:wave: INFO     This is SWE-agent version 1.0.1 (hash='9edc7dc1e9aa9dc06a75d9dce1ebd9472984462e') with SWE-ReX version 1.2.1 (rex_hash='unavailable').
:wrench: DEBUG    No .env file found
:person_running: DEBUG    Loading instances from SWEBenchInstances(subset='lite', split='dev', deployment=ModalDeploymentConfig(image='python:3.11', startup_timeout=180.0, runtime_timeout=60.0, deployment_timeout=1800.0,
            modal_sandbox_kwargs={}, type='modal', install_pipx=True), type='swe_bench', filter='.*', slice=':3', shuffle=False, evaluate=False)
Traceback (most recent call last):
  File "condaenv/bin/sweagent", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/storage/home/vseeker/workspace/swe-eval/SWE-agent/sweagent/run/run.py", line 91, in main
    run_batch_main(remaining_args)
  File "/storage/home/vseeker/workspace/swe-eval/SWE-agent/sweagent/run/run_batch.py", line 423, in run_from_cli
    run_from_config(BasicCLI(RunBatchConfig, help_text=help_text).get_config(args))  # type: ignore
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/storage/home/vseeker/workspace/swe-eval/SWE-agent/sweagent/run/run_batch.py", line 413, in run_from_config
    RunBatch.from_config(config).main()
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/storage/home/vseeker/workspace/swe-eval/SWE-agent/sweagent/run/run_batch.py", line 198, in from_config
    instances = config.instances.get_instance_configs()
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1023, opened 2025-03-12
run batch processing with modal

vsee:
The current version is unable to run batch executions on modal because Docker seems to be the default environment. This PR fixes that. Not sure if I am missing something though.

#1022

### DIFF
--- a/sweagent/run/batch_instances.py
+++ b/sweagent/run/batch_instances.py
@@ -10,7 +10,6 @@ from swerex.deployment.config import (
     DockerDeploymentConfig,
     DummyDeploymentConfig,
     LocalDeploymentConfig,
-    ModalDeploymentConfig,
 )
 from typing_extensions import Self
 
@@ -277,12 +276,6 @@ class SWEBenchInstances(BaseModel, AbstractInstanceSource):
 
         if isinstance(self.deployment, DockerDeploymentConfig):
             self.deployment.platform = "linux/amd64"
-
-        if not isinstance(self.deployment, DockerDeploymentConfig) and not isinstance(
-            self.deployment, ModalDeploymentConfig
-        ):
-            msg = "Only docker and modal backends are supported for batch runs."
-            raise ValueError(msg)
 
         instances = [
             SimpleBatchInstance.from_swe_bench(instance).to_full_batch_instance(self.deployment) for instance in ds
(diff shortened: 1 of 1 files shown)


sweagent/run/batch_instances.py
```diff
@@ -269,7 +274,16 @@ def get_instance_configs(self) -> list[BatchInstance]:
         from datasets import load_dataset
 
         ds: list[dict[str, Any]] = load_dataset(self._get_huggingface_name(), split=self.split)  # type: ignore
-        self.deployment.platform = "linux/amd64"
+
+        if isinstance(self.deployment, DockerDeploymentConfig):
+            self.deployment.platform = "linux/amd64"
+
+        if not isinstance(self.deployment, DockerDeploymentConfig) and not isinstance(
```
klieret, line 281: I'd rather not limit this here in case someone wants to generalize swe-rex etc.

klieret (approved):

klieret:
Thanks, I wasn't aware of that! Really appreciate the PR.

klieret:
I'll merge this, but I might actually also go ahead and remove these default settings for the docker deployment (`platform` and python standalone). I think they're currently just creating confusion. I'd rather extend the documentation for them.

sweagent/run/batch_instances.py
vsee: yes, good point

vsee:
> I'll merge this, but I might actually also go ahead and remove later these default settings for the docker deployment (`platform` and python standalone). I think they're currently just creating confusion. I'd rather extend the documentation for them.

@klieret Sounds good!

Currently the Swe-Rex [DockerDeployment](https://github.com/SWE-agent/SWE-ReX/blob/6a7b005cd8e7f21b0e580a09a1438c1f2abce492/src/swerex/deployment/docker.py#L147) packs a whole lot more into its image than simply the BASE_IMAGE this patch gives to the ModalDeployment. I don't know enough context to say whether that is a problem or not. So probably good if you doublecheck this.

klieret:
Shouldn't be a problem at all! This is all about how to run the tools:

1. With the `python_standalone_dir` (which is what this extra code is about): We add a whole different python installation to the image. The tools should then use that instead of the python that's also used for the tasks instance we're solving
2. Without it: Tools use the python from the task instance. Which is _usually_ fine, but just not always

Actually our benchmark results are currently still with option 2, so it works pretty well

vsee:
Ah, got it. Thanks.

ofirpress:
Thanks for making a PR!

Issue #1042, opened 2025-03-18
Retry run batch on existing traj cause keyerror


FRAOTIAC:
### Describe the bug

After a unfinished batch run, run it again will rise following error:

```
Traceback (most recent call last):
  File "/home/lucas/wsl2_workshop/SWE-agent/.venv/bin/sweagent", line 10, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/lucas/wsl2_workshop/SWE-agent/sweagent/run/run.py", line 91, in main
    run_batch_main(remaining_args)
  File "/home/lucas/wsl2_workshop/SWE-agent/sweagent/run/run_batch.py", line 426, in run_from_cli
    run_from_config(BasicCLI(RunBatchConfig, help_text=help_text).get_config(args))  # type: ignore
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/lucas/wsl2_workshop/SWE-agent/sweagent/run/run_batch.py", line 416, in run_from_config
    RunBatch.from_config(config).main()
  File "/home/lucas/wsl2_workshop/SWE-agent/sweagent/run/run_batch.py", line 243, in main
    self.main_single_worker()
  File "/home/lucas/wsl2_workshop/SWE-agent/sweagent/run/run_batch.py", line 261, in main_single_worker
    self.run_instance(instance)
  File "/home/lucas/wsl2_workshop/SWE-agent/sweagent/run/run_batch.py", line 299, in run_instance
    if self.should_skip(instance):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/lucas/wsl2_workshop/SWE-agent/sweagent/run/run_batch.py", line 390, in should_skip
    exit_status = data["info"].get("exit_status", None)
                  ~~~~^^^^^^^^
KeyError: 'info'
```

### Steps/commands/code to Reproduce

```
sweagent run-batch --num_workers=1 \
    --instances.type=swe_bench --instances.subset=lite --instances.split=test \
    --instances.shuffle=True --instances.evaluate=True \
    --instances.deployment.docker_args=--memory=10g --config config/sweagent_heavy_sbl_250212.yaml \
    --instances.slice :5
```

Canceling the first run before it finished and rerun will reporduce this error.

### Error message/results

```
bash scripts/run-batch.bash
[...]

PR #1043, opened 2025-03-18
[#1042] fix(run_batch): handle JSON parsing errors in trajectory check

FRAOTIAC:
- Added try-except block to handle JSON parsing errors when checking existing trajectories.
 - Logs an error message and removes the trajectory if parsing fails.

 Resolves: #1042


klieret (approved):
thanks!

PR #1050, opened 2025-03-23
chore: Change function calling check Exception to only raising a warning

### LINKED ISSUE #1038: Allow skipping function calling verification with LiteLLM
### Describe the feature

Because certain deployment names e.g. `azure/4o-mini` might differ from the actual names, it should be possible to skip the function calling verification step / override it.

I am unsure if a mapping is possible (https://docs.litellm.ai/docs/completion/model_alias), this change however would mainly make SWEAgent work for deployments with weird names (which sadly exist for whatsoever reason).

### Potential Solutions

Potential solution looks like this:

```python
if tools.use_function_calling:
            if not litellm.utils.supports_function_calling(model=self.config.name) and os.getenv("CHECK_FUNCTION_CALLING", "false").lower() == "true":
                msg = (
                    f"Model {self.config.name} does not support function calling. If your model"
                    " does not support function calling, you can use `parse_function='thought_action'` instead. "
                    "See https://swe-agent.com/latest/faq/ for more information."
                )
                raise ModelConfigurationError(msg)
```

jpaodev:
#### Reference Issues/PRs
Fixes #1038 

This Merge Request removes the hard check for specific model names. It changes the following lines:

```python
if tools.use_function_calling:
            if not litellm.utils.supports_function_calling(model=self.config.name):
                msg = (
                    f"Model {self.config.name} does not support function calling. If your model"
                    " does not support function calling, you can use `parse_function='thought_action'` instead. "
                    "See https://swe-agent.com/latest/faq/ for more information."
                )
                raise ModelConfigurationError(msg)
```
This current code does not allow for custom model names, which are (sadly) sometimes the case (e.g. `azure/4o-mini` would not work, because it's not in the list of LiteLLM models)

Because sometimes custom names / mappings are required, this change turns the Exception into a logger warning instead, so the user can see the risk, however still use the software if it's required to be used with a custom name.

### DIFF
--- a/sweagent/agent/models.py
+++ b/sweagent/agent/models.py
@@ -560,6 +560,8 @@ class LiteLLMModel(AbstractModel):
         self.config: GenericAPIModelConfig = args.model_copy(deep=True)
         self.stats = InstanceStats()
         self.tools = tools
+        self.logger = get_logger("swea-lm", emoji="🤖")
+        
         if tools.use_function_calling:
             if not litellm.utils.supports_function_calling(model=self.config.name):
                 msg = (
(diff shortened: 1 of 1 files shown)


klieret (approved):
Thanks

PR #1054, opened 2025-03-24
Use default trajectories dir according to ENV

vsee:
According to the changes in #1046 where the default trajectories directory is set here:

https://github.com/SWE-agent/SWE-agent/blob/71f2ed93d544b6fbd3b5d52792679579175fc881/sweagent/__init__.py#L46

It looks like it should be used here:

https://github.com/SWE-agent/SWE-agent/blob/71f2ed93d544b6fbd3b5d52792679579175fc881/sweagent/run/run_batch.py#L115


klieret (approved):
Thanks for the catch, appreciate it!

PR #1075, opened 2025-04-15
chore: Update to Python 3.13

reneleonhardt:
#### Updates
- Update image references to python:3.13-slim
- Add Python 3.13 to pytest
- Let Dependabot update NodeJS

#### Fixes
- curl isn't installed by default in Debian (`FROM python`)
- `docker build` example was missing the context (first positional argument)

#### Documentation errors
Markdown example `sweagent run --env.image swe-agent-tiny` results in
> SettingsError
> error parsing CLI: unrecognized arguments: --env.image swe-agent-tiny

pydoc example `--env.docker.image` doesn't work either.

#### TODO
- Dependabot should also update Dockerfiles and docker-compose
- Docker base images `ubuntu:14.04` haven't been built for three years
- release_dockerhub.sh references `-f docker/swe.Dockerfile` which doesn't exist
- pytest coverage only runs for `if: matrix.python-version == '3.11'`, maybe that could be 3.13 too?

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,3 +5,7 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+  - package-ecosystem: "npm" # Update package.json
+    directory: "/sweagent/frontend"
+    schedule:
+      interval: "weekly"
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -29,7 +29,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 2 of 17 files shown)


klieret:
Hi @reneleonhardt , I appreciate the work on this, but it's much better practice to open separate PRs for differen things.

For example, I want to keep the py3.11 for now, and don't want the node update thing because it just generates additional noise (the web interface to swe-agent is currently not properly supported anyway and we might perhaps remove it completely)

klieret:
Happy to merge the doc fixes with the docker command if you open a separate PR.

klieret:
But closing this one because it's too many changes at once

PR #1100, opened 2025-05-04
Fix trajectory messages to reflect what LLM actually saw

celestinoalan:
## Fix: Ensure Trajectory Messages Accurately Reflect LLM Input/Output

**Problem:**

Previously, the `messages` field saved within each `TrajectoryStep` in the `.traj` file did not accurately represent the input the language model received for that specific step.

This occurred because:
1.  The `DefaultAgent.step` method first called the LLM with `self.messages` (which applies history processors).
2.  It then updated the raw `self.history` with the LLM's response (assistant message) and the resulting observation (user/tool message).
3.  Finally, it called `self.add_step_to_trajectory`, which accessed `self.messages` *again*. This re-applied the history processors to the *updated* history, leading to a different set of messages being saved than what the LLM originally saw (e.g., with `LastNObservations`, the observation from the *current* step was included, while the one actually sent to the LLM was potentially omitted).

**Solution:**

This PR introduces the following changes to `sweagent/agent/agents.py` to fix this discrepancy and improve usability for fine-tuning:

1.  **Capture LLM Input:** In `DefaultAgent.step`, the result of `self.messages` is now captured *before* the call to `forward_with_handling`. This variable (`messages_for_llm`) accurately stores the processed messages sent to the LLM.
2.  **Store Full Turn in Trajectory:**
    *   After `add_step_to_history` adds the raw assistant response and user/tool observation to `self.history`, a new list `messages_for_trajectory` is created.
    *   This list combines the captured `messages_for_llm` with the last two entries from the raw `self.history` to further facilitate fine-tuning later.
    *   `add_step_to_trajectory` is modified to accept this list and save it in the `TrajectoryStep`.

**Outcome:**

The `messages` field in each trajectory step now accurately represents the full conversation turn for that step:
- The processed input messages exactly as seen by the LLM.
- The raw, unprocessed assistant message (LLM response).
- The raw, unprocessed user/tool message (observation).

This provides a correct record of the interaction for analysis and makes the trajectory format more convenient for fine-tuning, as the input and target completion are readily available within the step's `messages` without needing to reconstruct the state or worry about the effect of history processors on the output messages.

## Reproducing the Original Issue

To observe the discrepancy in the trajectory messages before this fix, follow these steps:

1.  **Modify Configuration (`config/function_calling.yaml`):**
    *   Set the `LastNObservations` history processor to keep only the last observation by changing `n` to `1`:
        ```yaml
        history_processors:
          - type: last_n_observations
            n: 1 # <-- Change this from the default
        ```
    *   For simplicity and faster reproduction, remove or comment out all the sections related to demonstrations within the `templates` block.

2.  **Run SWE-agent:** Execute the following command from the root of the repository:
    ```bash
    sweagent run --config ./config/function_calling.yaml \
      --agent.model.name=gpt-4o \
      --agent.model.per_instance_cost_limit=2.00 \
      --env.repo.github_url=https://github.com/SWE-agent/test-repo \
      --problem_statement.github_url=https://github.com/SWE-agent/test-repo/issues/1 \
      --output_dir=./generated_trajectories
    ```

3.  **Inspect Trajectory:**
    *   Once the run completes, open the generated `.traj` file located in `./generated_trajectories/`. The filename will correspond to the issue ID (e.g., `test-repo__test-repo-1.traj`).
    *   Examine the `trajectory` list within the JSON structure. Look specifically at the `messages` field of the *last* step in the trajectory.
    *   **Observation (Original Bug):** You would notice that the last message in this `messages` list is the user/tool observation corresponding to the *current* step. Because `LastNObservations` was set to `n=1` and applied *after* the history was updated, the previous observation, which is the one that was *actually* sent to the LLM (from the *previous* step), is omitted. This `messages` list does not accurately reflect the input the LLM received for this final step.

### DIFF
--- a/sweagent/agent/agents.py
+++ b/sweagent/agent/agents.py
@@ -972,6 +972,10 @@ class DefaultAgent(AbstractAgent):
         # attributes (e.g., if we want to requery the model for a bash syntax error, we
         # need to have the previous model output to format the requery template)
         step = StepOutput()
+        # Record the exact messages that are sent to the language model for this call.
+        # Do this *before* the actual model call, so if the call or parsing fails,
+        # the step object attached to the exception still has the correct input.
+        step.input_messages = copy.deepcopy(history)
         try:
             # Forward model and get actions
             self._chook.on_model_query(messages=history, agent=self.name)
@@ -1036,6 +1040,7 @@ class DefaultAgent(AbstractAgent):
             """Requeries the model if the error is a format/blocklist/bash syntax error."""
             self.logger.warning("Requerying model after %s (%dth requery)", type(exception).__name__, n_requeries)
             step: StepOutput = getattr(exception, "step", StepOutput())
+            # Pass the history that led to the error to the trajectory step
             self.add_step_to_trajectory(step)
             exception_message = getattr(exception, "message", "")
             if not exception_message:
@@ -1163,6 +1168,13 @@ class DefaultAgent(AbstractAgent):
         )
 
     def add_step_to_trajectory(self, step: StepOutput) -> None:
+        """Add a step to the trajectory.
+
+        The ``messages`` field in the trajectory step will contain exactly the
+        input messages provided to the language model for this attempt (stored
+        in ``step.input_messages``).
+        """
+
         trajectory_step = TrajectoryStep(
             {
                 "action": step.action,
@@ -1171,7 +1183,7 @@ class DefaultAgent(AbstractAgent):
                 "thought": step.thought,
                 "execution_time": step.execution_time,
                 "state": step.state,
-                "messages": self.messages,
+                "messages": step.input_messages,  # Always use the recorded input messages
                 "extra_info": step.extra_info,
             },
         )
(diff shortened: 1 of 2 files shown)


klieret:
It's true, this currently incorrectly implemented, however, I think the fix here still leaves problems (we should directly set `step` when we query the model).

klieret:
I'll probably close this and add my own fix

alancelestino:
Hey @klieret,

Thanks for pointing that out! I just changed the approach considerably:

1. `StepOutput` now has an `input_messages` field only containing messages that were used as the input to the LLM (this is set in `self.forward`, immediately before making the LLM call).
2. `self.add_step_to_trajectory` only adds `step.input_messages` (still in the same positions in the code).

Do you see any issues with this approach, @klieret? I'll be using this code for fine-tuning soon so I'm eager to get that out of the way 😅 

Cheers

klieret:
I meant to add it here: https://github.com/SWE-agent/SWE-agent/blob/bb80cbe3d48c6d1f3140ebe20dc52bb4408a46a4/sweagent/agent/agents.py#L990 , because this is exactly what we pass on to the query and just add it as step.messages

I might be able to do it later today, but traveling at the moment. But your solution will also already be correct in 99% of the cases.

klieret:
#1107 does the change as I described it. It's not merged, because it will need a bit of doc updates on top

alancelestino:
Hey @klieret,

Cool, that is working as expected on my end :) Thank you!

Cheers

klieret:
Closing in favor of #1107

PR #1132, opened 2025-05-16
feat: support fully local runs via repo_base_dir and env_dir environment variable

askender:
#### Reference Issues/PRs
Fixes #1234   Support the feature request issue

#### What does this implement/fix?
The first commit adds support for two environment variables:
```
env_dir = os.environ.get('ENV_DIR', '/root').rstrip('/')
repo_base_dir = os.environ.get('REPO_BASE_DIR', '').rstrip('/')
```
This replaces hardcoded paths with configurable values.

The second commit includes a small but useful patch (4 lines of code) related to this change.
The third commit addresses a typechange bug (optional but recommended).

Finally, the fourth commit enables full support for running swe-agent locally without Docker.

Please be cautious when running the code outside of Docker — this setup assumes a safe and isolated environment.

Hope this helps others who need local execution.
Have fun hacking!


askender:
I forgot to run `ruff` in the previous commits.
I've now added a 5th commit to clean up the code and ensure it's more Pythonic.
You're welcome to cherry-pick this PR if it helps.

klieret:
Have you tested this? Pretty sure that this doesn't work like this without making sure the env variables are set in the agent config

sweagent/agent/agents.py
```diff
@@ -56,6 +57,8 @@
 from sweagent.utils.log import get_logger
 from sweagent.utils.patch_formatter import PatchFormatter
 
+env_dir = os.environ.get('ENV_DIR', '/root').rstrip('/')
+repo_base_dir = os.environ.get('REPO_BASE_DIR', '').rstrip('/')
```
klieret, line 61: Isn't that just what we set as `$ROOT`?

askender:
> What workflow are you using this for / how are you using this / have you tested this?

Yes, I’ve tested it on macOS without Docker.
My goal is to run it on a clean Linux environment, which is safe even without Docker.
Since I need to test a large number of projects (e.g., GitHub repos), performance is important — the faster, the better.

This workflow allows me to debug and add new features more clearly than using Docker.

> Doesn’t this still start Docker unless you switch out the deployment to a local deployment config?

Yes, but I’ve already set the deployment type to `"local"`.
Here’s the script I use:

```bash
#!/bin/bash

export ENV_DIR=/tmp/root
export REPO_BASE_DIR=/tmp/repos

DEPLOYMENT_TYPE="local"

while read -r repo instance_id base_commit environment_setup_commit; do
  echo "Processing instance: $instance_id"

  owner_repo=$(echo "$repo" | tr -d '"')
  repo_name=$(basename "$owner_repo")

  sweagent run \
    --agent.model.name "$MODEL_NAME" \
    --agent.model.api_base "$API_BASE" \
    --agent.model.per_instance_cost_limit "$COST_LIMIT" \
    --agent.model.total_cost_limit 0 \
    --env.deployment.type "$DEPLOYMENT_TYPE" \
    --env.repo.path "./repos/$repo_name" \
    --env.repo.base_commit "$base_commit" \
    --problem_statement.path "./instances_lite_test/$instance_id.md"

done < issues_list.txt
```

This supports both `local` and `docker` deployment modes.

I’m not using `sweagent run-batch` because the above simple script works better for my needs — I can reuse the same directory to test multiple issue instances without resetting everything each time.

askender:
> What workflow are you using this for / how are you using this / have you tested this?

Yes, I’ve tested it on macOS without Docker.
My goal is to run it on a clean Linux environment, which is safe even without Docker.
Since I need to test a large number of projects (e.g., GitHub repos), performance is important — the faster, the better.

This workflow allows me to debug and add new features more clearly than using Docker.

> Doesn’t this still start Docker unless you switch out the deployment to a local deployment config?

Yes, but I’ve already set the deployment type to `"local"`.
Here’s the script I use:

```bash
#!/bin/bash

export ENV_DIR=/tmp/root
export REPO_BASE_DIR=/tmp/repos

DEPLOYMENT_TYPE="local"

while read -r repo instance_id base_commit environment_setup_commit; do
  echo "Processing instance: $instance_id"

  owner_repo=$(echo "$repo" | tr -d '"')
  repo_name=$(basename "$owner_repo")

  sweagent run \
    --agent.model.name "$MODEL_NAME" \
    --agent.model.api_base "$API_BASE" \
    --agent.model.per_instance_cost_limit "$COST_LIMIT" \
    --agent.model.total_cost_limit 0 \
    --env.deployment.type "$DEPLOYMENT_TYPE" \
    --env.repo.path "./repos/$repo_name" \
    --env.repo.base_commit "$base_commit" \
    --problem_statement.path "./instances_lite_test/$instance_id.md"

done < issues_list.txt
```

This supports both `local` and `docker` deployment modes.

I’m not using `sweagent run-batch` because the above simple script works better for my needs — I can reuse the same directory to test multiple issue instances without resetting everything each time.

klieret:
I'll take a look at this very soon. We've actually been talking a little bit about making swe-agent more interactive and this fits in with that

sweagent/environment/swe_env.py
```diff
@@ -142,6 +145,10 @@ def reset(self):
             info: additional information (e.g. debugging information)
         """
         self.communicate(input="cd /", check="raise")
+        # Remove the tools directory if it's under /tmp to ensure a clean reset.
+        # Deleting from /tmp is safe and avoids leftover state from previous runs.
+        if env_dir.startswith('/tmp'):
+            self.communicate(f"rm -rf {env_dir}/tools", check="raise")
```
klieret, line 151: This will be confusing behavior

sweagent/environment/swe_env.py
```diff
@@ -153,10 +160,11 @@ def _reset_repository(self) -> None:
             # todo: Currently has swe-ft specific change: The original repo.copy isn't called, because the repo is already
             # present. However, reset --hard <BRANCH> also doesn't work. So modified it here to do a checkout instead.
             startup_commands = [
-                f"cd /{self.repo.repo_name}",
+                f"cd {repo_base_dir}/{self.repo.repo_name}",
                 "export ROOT=$(pwd -P)",
                 "git status",
-                "git fetch",
+                # "git fetch", # it seems useless
+                "git reset --hard", # solve typechange
```
klieret, line 167: We cannot do `git reset --hard`. Would break certain swe-bench instances, for example

askender:
You're right. I've retained only the two essential commits.
Additionally, I've updated the first commit to pass the `ruff check`.

sweagent/agent/agents.py
```diff
@@ -812,13 +816,14 @@ def attempt_autosubmission_after_error(self, step: StepOutput) -> StepOutput:
         repo_name = "/"
         if self._env.repo is not None:
             repo_name = f"/{self._env.repo.repo_name}"
-        submission_command = "git add -A && git diff --cached > /root/model.patch"
+        repo_name = f"{repo_base_dir}{repo_name}"
```
klieret, line 819: This feels wrong, isn't there a `/` missing here? 

Also probably in line 816 should then set this to `repo_base_dir` instead of `/`

sweagent/environment/swe_env.py
```diff
@@ -216,8 +220,13 @@ def communicate(
         """
         self.logger.log(logging.TRACE, "Input:\n%s", input)  # type: ignore
         rex_check = "silent" if check else "ignore"
+
+        bash_lines = [f'export ENV_DIR="{env_dir}"', f'export REPO_BASE_DIR="{repo_base_dir}"', input]
```
klieret, line 224: Oh I hadn't noticed that this was in `communicate`. This should definitely be moved to some startup commands, rather than being appended to every single communicate call.

klieret (changes requested):
The other thing is, wouldn't it make more sense to just set these two variables as part of the environment config? Currently, we only need them in `Agent`, `SWEEnv` and the tools class, so they all have access tot he config setting.

Then, we set them as env variables to expose them to the tools.

klieret:
There's also still at least one test that is failing, did you look into why that happens? 

It's best to run ptyest with

```
-k name_of_test -s --capture=no --log-cli-level=DEBUG
```

to see all agent log messages

sweagent/agent/agents.py
askender: env_dir = os.environ.get('ENV_DIR', '/root').rstrip('/')
repo_base_dir = os.environ.get('REPO_BASE_DIR', '').rstrip('/')

refer to
```
env_dir = os.environ.get('ENV_DIR', '/root').rstrip('/')
repo_base_dir = os.environ.get('REPO_BASE_DIR', '').rstrip('/')
```

for the default config:
env_dir = '/root'
repo_base_dir = ''

I didn't change 816-818
```
        repo_name = "/"
        if self._env.repo is not None:
            repo_name = f"/{self._env.repo.repo_name}"
```

In this patch:
819 chage from:
```
submission_command = "git add -A && git diff --cached > /root/model.patch"
```
to:
```
repo_name = f"{repo_base_dir}{repo_name}"
submission_command = f"git add -A && git diff --cached > {env_dir}/model.patch"
```

So, in default config,
repo_name should be same because `repo_base_dir=''`
If people change them to sth like:
env_dir = '/tmp/root'
repo_base_dir = '/tmp/repos'

```
repo_name = /the_repo_name
repo_name = f"/tmp/repos{repo_name}"
`

So, there is not a `/` missing here.

---

You suggest:
change the code from
```
repo_name = "/"
if self._env.repo is not None:
    repo_name = f"/{self._env.repo.repo_name}"
repo_name = f"{repo_base_dir}{repo_name}"
```
to
```
repo_name = "{repo_base_dir}"
if not repo_name:
    repo_name = "/"
if self._env.repo is not None:
    repo_name = f"{repo_name}/{self._env.repo.repo_name}"
```
which has the same result.

my raw patch add 1 new line.
Your suggestion changed 2 lines, add 2 new lines.

---
You may suggest that `repo_base_dir` should start with a `/` at all times, as it is clearer.
```
repo_base_dir = os.environ.get('REPO_BASE_DIR', '').rstrip('/') # rstrip is not so good for '/'
if repo_base_dir == "/":
    repo_base_dir = "/"
```

I will do that when `add these two variables as part of the environment config` as you said.

So: 

raw code:
```
PurePosixPath("/") / self._env.repo.repo_name / path
 f"mkdir /{self.repo_name}",
```
patch_not_so_good:
```
PurePosixPath(f"{repo_base_dir}/") / self._env.repo.repo_name / path
f"mkdir {repo_base_dir}/{self.repo_name}",
```
patch_as_you_suggest:
```
PurePosixPath(self._env.repo.repo_base_dir) / self._env.repo.repo_name / path
f"mkdir PurePosixPath(self.repo_base_dir) / self.repo_name",
```

Thanks a lot!

```
When I write code, I find myself standing at a forked path:
One way is **lazy** — swift and brief, a whisper of logic that gets the job done.
The other, **semantic** — deliberate, clear, a language of meaning stretched in lines.
And I ask myself:
Why does my hand so often reach for the shortcut,
when my mind knows the beauty lies in clarity?
```

sweagent/environment/swe_env.py
askender: Sure, I’ll do that if the new PR is worthwhile.

reneleonhardt:
Is it possible to improve the documentation to show the fastest way for a fully local run?
https://swe-agent.com/latest/usage/coding_challenges/ shows some bits and pieces, maybe there could be a dedicated guide for a fully local run with ollama with a current LLM and without accessing GitHub.

FYI the README is outdated, GPT-4o and Claude Sonnet 3.7 are not the latest SWE-bench verified models.

klieret:
Ah shoot, i realized that me renaming the configs introduced merge conflicts. Happy to resolve them, but also realized I don't have permissions to push on this branch. Can you give me access like here: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork ?

reneleonhardt:
@askender Can you allow maintainers to push to your branch?

askender:
After becoming the admin of [https://github.com/VoyahAGI/SWE-agent](https://github.com/VoyahAGI/SWE-agent), I noticed that the **"Allow edits by maintainers"** option still doesn't appear, which is quite strange.
To work around this, I added **klieret** as a maintainer of the repository.

klieret:
Thanks for this! However, in the end I think mini-swe-agent just is the better project at this point for doing fully local runs

PR #1140, opened 2025-05-19
FIX: Use correct local model tokenizer for computing token consumption

Liqs-v2:
## Issue this PR addresses
SWE-agent tracks the token consumption of the agent using the LiteLLM `token_counter`. However, this does not work as expected for local models.

### Expected
The LiteLLM functionality for tokenizing and counting tokens with the model specific tokenizer is utilized.
### Actual
The `token_counter` is invoked as is. This results in using the OpenAI tokenizer to count tokens (default fallback), resulting in potentially incorrect token counts.
### Why this is an issue?
As SWE-agent is a project that is popular in the research community, and this community is moving towards training local models. In an effort of scientific rigor, we should use accurate token consumption statistics.

## Changes this PR introduces
If the configuration is in line with that of a local model (cost limits are 0), we set up the model-specific tokenizer using HF `transformers` and `AutoTokenizer`. Due to a bug in LiteLLM, we pass the `backend_tokenizer`. We update the `token_counter` to use this `custom_tokenizer` if available.

### DIFF
--- a/sweagent/agent/models.py
+++ b/sweagent/agent/models.py
@@ -24,7 +24,6 @@ from tenacity import (
     stop_after_attempt,
     wait_random_exponential,
 )
-from transformers import AutoTokenizer
 
 from sweagent import REPO_ROOT
 from sweagent.exceptions import (
@@ -599,6 +598,7 @@ class LiteLLMModel(AbstractModel):
         self.lm_provider = litellm.model_cost.get(self.config.name, {}).get("litellm_provider", self.config.name)
         if self.config.per_instance_cost_limit == 0 and self.config.total_cost_limit == 0:  # Local model
             if "/" in self.lm_provider:
+                from transformers import AutoTokenizer
                 self.custom_tokenizer = {}
                 self.custom_tokenizer["provider"] = self.lm_provider.split("/")[0]
 
--- a/sweagent/agent/models.py
+++ b/sweagent/agent/models.py
@@ -599,6 +599,7 @@ class LiteLLMModel(AbstractModel):
         if self.config.per_instance_cost_limit == 0 and self.config.total_cost_limit == 0:  # Local model
             if "/" in self.lm_provider:
                 from transformers import AutoTokenizer
+
                 self.custom_tokenizer = {}
                 self.custom_tokenizer["provider"] = self.lm_provider.split("/")[0]
 
--- a/sweagent/agent/models.py
+++ b/sweagent/agent/models.py
@@ -11,6 +11,7 @@ from abc import ABC, abstractmethod
 from pathlib import Path
 from threading import Lock
 from typing import Annotated, Any, Literal
+from transformers import AutoTokenizer
 
 import litellm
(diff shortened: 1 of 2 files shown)


sweagent/agent/models.py
```diff
@@ -594,7 +596,28 @@ def __init__(self, args: GenericAPIModelConfig, tools: ToolConfig):
...
+
+                if self.custom_tokenizer["provider"] not in litellm.provider_list:
+                    self.logger.warning(
+                        f"Local model {self.lm_provider} not found in LiteLLM provider list. Using default tokenizer."
+                    )
+                    self.custom_tokenizer = None
+                else:
+                    self.custom_tokenizer["identifier"] = "/".join(self.lm_provider.split("/")[1:])
+                    self.custom_tokenizer["type"] = "huggingface_tokenizer"
+                    # Use backend tokenizer as workaround for litellm HF tokenizer bug
```
klieret, line 613: what bug are you referring to here? Do you have an issue link?

klieret (changes requested):
Thank you for opening this PR!

I have to think about this, but it seems to me that a cleaner way to do this might be to have the user specifically configure the tokenizer as part of the config (?). 

Right now this has a lot of ifs and elses and adds a big new dependency.

My ideal solution would be to stick to how litellm wants to solve this and not try and take care of this ourselves or if there's edge cases this doesn't cover to solve it with explicit user options.

sweagent/agent/models.py
Liqs-v2: No, I couldn't find an issue directly addressing this in LiteLLM and was also considering fixing it in LiteLLM if I find the time. It's really just adding an if-branch. I'll open an issue later.

Basically, [their token counter implementation](https://github.com/BerriAI/litellm/blob/a676f69cab1e7b7a21c9b162fab6968bdc96bcd4/litellm/litellm_core_utils/token_counter.py#L510) expects a `Tokenizer` that returns an [Encoding](https://huggingface.co/docs/tokenizers/api/encoding), because they access the `ids` field after calling the tokenizer. However, this is quite unintuitive, because the typical way one would set up a tokenizer is with `AutoTokenizer` which does not return an `Encoding` but a list of ids directly.

What the fix in LiteLLm would be imo is:
```python
def count_tokens(text: str) -> int:
  enc = tokenizer_json["tokenizer"].encode(text)
  if isinstance(enc, list):
      return len(enc)
  elif hasattr(enc, "ids"):
      return len(enc.ids)
  else:
      raise TypeError(f"Unexpected return type from encode: {type(enc)}")
```

klieret:
Hi @Liqs-v2 , thanks again for this! I've squashed these commits and then somewhat simplified this in #1224 (so you still retain your authorship on the squashed commit ;) ). This doesn't have your workaround with the `AutoTokenizers`, but I think it should get 90% there. Hopefully this still makes everything possible, even if slightly less convenient. Open for further discussion, but ideally maybe something to bring up with `litellm` directly.

PR #1144, opened 2025-05-21
CHANGE: Rename configs

klieret:


klieret:
Further rename to just have a `default.yaml`, `default_no_fcalls.yaml` on base level

PR #1178, opened 2025-05-30
Fixes #1166: strip incorrect for XML function calling

### LINKED ISSUE #1166: Striping for XML Function calling parser
### Describe the bug

Hey ! 

The `XMLFunctionCallingParser` is striping arguments from the models when converting to tool call, which can lead into issues when using old_str/new_str tools, as the fields would be modified. For instance, if the model wants to replace `\n   a`, that would actually be seen as `a`. One option would be to use instead `re.sub(r"^\n|\n$", "", text)` to account for new lines in XML. Curious if that's an intentional choice you made to strip before creating a PR to fix.

Thanks in advance

### Steps/commands/code to Reproduce

Agent workflows with XMLFunctionCallingParser

### Error message/results

N/A

### System Information

N/A

### Checklist

- [x] I'm running with the latest docker container/on the latest development version (i.e., I ran `git pull`))
- [x] I have enclosed code/log messages in triple backticks ([docs](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#quoting-code)) and clicked "Preview" to make sure it's displayed correctly.

Gguinet:
The `XMLFunctionCallingParser` is striping arguments from the models when converting to tool call, which can lead into issues when using old_str/new_str tools, as the fields would be modified. 

For instance, if the model wants to replace \n   a, that would actually be seen as a. 
This PR uses re.sub(r"^\n|\n$", "", text) removes at most one newline at begging and end of the field, such that 
```
<parameter=example_parameter_2>
This is the value for the second parameter
that can span
multiple lines
</parameter>
</function>
```
gives as param 
```
This is the value for the second parameter
that can span
multiple lines
```

### DIFF
--- a/sweagent/tools/parsing.py
+++ b/sweagent/tools/parsing.py
@@ -276,7 +276,11 @@ class XMLFunctionCallingParser(AbstractParseFunction, BaseModel):
             msg = f"Command '{fn_name}' not found in list of available commands."
             raise FormatError(msg)
 
-        params_dict = {param[0]: param[1].strip() for param in re.findall(FN_PARAM_REGEX_PATTERN, fn_body, re.DOTALL)}
+        params_dict = {
+            param[0]: re.sub(r"^\n|\n$", "", param[1])
+            for param in re.findall(FN_PARAM_REGEX_PATTERN, fn_body, re.DOTALL)
+        }
+
         if "view_range" in params_dict:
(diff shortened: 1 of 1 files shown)


klieret:
Thanks again for this contribution, really valued!

PR #1181, opened 2025-05-31
Doc: Recommend sonnet 4 as the default

klieret:


SUmidcyber:
İmport update , so we are looks tools

PR #1199, opened 2025-06-11
Fix GenericAPIModelConfig.id's property to avoid casting errors on valid None types

carlosejimenez:
…

For `GenericAPIModelConfig`'s `id` property, we have to manually cast `top_p` to a string, as it can be None.
We additionally remove directory nesting from happening automatically in model names with forward slashes (e.g. `openrouter/model-name`).

In general though, we may want to discuss the right policy for handling this case, and changes to the default `id` property.


sweagent/agent/models.py
```diff
@@ -180,7 +180,14 @@ def choose_api_key(self) -> str | None:
 
     @property
     def id(self) -> str:
-        return f"{self.name}__t-{self.temperature:.2f}__p-{self.top_p:.2f}__c-{self.per_instance_cost_limit:.2f}"
+        name = self.name.replace("/", "--")
```
klieret, line 183: ouff, never thought about how this could give us a new directory 😓

sweagent/agent/models.py
```diff
@@ -180,7 +180,14 @@ def choose_api_key(self) -> str | None:
...
+        if self.top_p is not None:
+            top_p = f"{self.top_p:.2f}"
+        else:
+            top_p = "None"
```
klieret, line 187: wouldn't it be more consistent to just set it to 1.0 ?

klieret (approved):

PR #1214, opened 2025-06-19
Update models.py

x14ngch3n:
fix the format error when using human_thought mode. The action in code block should be the command itself, not the message dict

Extract the "message" field from the return dict of self._query()


x14ngch3n:
Has anyone reviewed it yet? I think it's important to the human thought model.

klieret (approved):
You're right, that wouldn't work right now! Thanks for spotting this!

PR #1223, opened 2025-06-25
[tmp] Enh/read images

klieret:


tools/sweer/bin/click_mouse
```diff
@@ -0,0 +1,41 @@
+#!/root/python3.11/bin/python3
```
klieret, line 1: the shebangs are currently mixed. Some say #!/usr/bin/env python3, others do the /usr/bin/env

PR #1236, opened 2025-07-07
Update docker-in-docker feature to specific version

davides93:
Switch the docker-in-docker feature to use a specific version for improved stability and compatibility.


klieret:
Unless there's known issues, I think I'd rather leave it as-is and fix things if they break.

PR #1238, opened 2025-07-09
Manage tool execution_timeout in the config

0xba1a:


john-b-yang:
Super neat, thanks for the PR @0xba1a! I think it looks great. Requested review from @klieret as well.

klieret:
Hi, decided to keep the scope of tools more limited in the end.

PR #1244, opened 2025-07-15
Add an option for the submit tool to submit a file

0xba1a:
As of now the submission tool generates a patch from the `git diff`. Other than that, if the user wants to get something else from the repo, they can indirectly do it by creating it as a new file in the repo and make it part of the patch.

Instead, this change enables submit tool to handle a file argument. If provided, that file will be simply submitted back to the orchestrator without creating a patch.

### DIFF
--- a/config/test/submit_file_test.yaml
+++ b/config/test/submit_file_test.yaml
@@ -1,0 +1,152 @@
+# Formerly called: anthropic_filemap.yaml
+# This template is heavily inspired by anthropic's computer use demo, but you can use
+# it with any LM.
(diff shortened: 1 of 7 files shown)


0xba1a:
It seems that `Check Markdown links` check failure is not related. It can be addressed in a separate PR.

klieret:
I get the motivation for this. Happy to link to this commit in the docs among extensions. However, we decided to largely keep the current state of swe-agent while most of our current efforts have moved to mini-swe-agent

PR #1263, opened 2025-08-01
Add link to ex9-fyi's multiagent fork

ex9-fyi:
Reference: https://github.com/SWE-agent/mini-swe-agent/pull/328


klieret:
Hi @ex9-fyi let me add links to interesting forks in the mini docs today (these are the swe-agent docs)

PR #1291, opened 2025-08-29
use docker.io prefix to pull image (for podman)

rambip:
#### Reference Issues/PRs

Fixes https://github.com/SWE-agent/SWE-agent/issues/1290

I edited every image name that was not specific enough with adding `docker.io/` prefix. It should now work with podman.


klieret (approved):
thanks!

PR #1322, opened 2025-12-19
Add ACP server command

luca-ctx:
Adds an ACP stdio server command, session mapping, and Zed ACP setup docs.


klieret:
Thanks, appreciate the suggestion, but we decided to keep the scope of this project focused on being a research project and limit integrations that we need to keep maintained

PR #1334, opened 2026-01-29
fix(run_batch): skip instances with uncaught dockerpull exceptions from previous…

HaotianChen616:
#### Reference Issues/PRs

None (discovered during local ARM testing)

**Problem:** When running batch jobs, instances that fail with infrastructure errors (e.g., `DockerPullError` when the Docker image doesn't exist for the current platform) are retried on every subsequent run. This happens because these failures occur before a trajectory file is created, so the existing skip logic in `_should_skip()` doesn't detect them.

This is particularly problematic on ARM machines (e.g., Apple Silicon Macs), where many x86-only images are unavailable.

**Solution:** Enhanced `_should_skip()` in `sweagent/run/run_batch.py` to also check `run_batch_exit_statuses.yaml` for instances with uncaught exceptions. Instances that previously failed with errors like `Uncaught DockerPullError` are now skipped on re-runs.

**Changes:**
- Refactored trajectory check logic for clarity
- Added yaml report parsing to detect instances with previous uncaught exceptions
- These infrastructure/platform errors won't resolve by retrying, so skipping is appropriate

### DIFF
--- a/sweagent/run/run_batch.py
+++ b/sweagent/run/run_batch.py
@@ -380,33 +380,53 @@ class RunBatch:
         if self._redo_existing:
             return False
 
+        instance_id = instance.problem_statement.id
         # Check if there's an existing trajectory for this instance
-        log_path = self.output_dir / instance.problem_statement.id / (instance.problem_statement.id + ".traj")
-        if not log_path.exists():
-            return False
+        log_path = self.output_dir / instance_id / (instance_id + ".traj")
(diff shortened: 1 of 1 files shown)


klieret:
Hi @zeus-cht , thanks for this and I get where you're coming from. However, my preference is to not complicate the codebase any more. To me this is more of a niche case.

PR #1336, opened 2026-02-05
fix: correct inverted logic in blocklist check

### LINKED ISSUE #1316: Inverted logic in ToolHandler.should_block_action blocklist check
### Description

In `sweagent/tools/tools.py`, the `ToolHandler.should_block_action` method has an inverted condition when checking the blocklist. The current implementation checks if blocklist items start with the action, instead of checking if the action starts with blocklist items.

### Location

`sweagent/tools/tools.py`  
Line: 385

### Current Code (Incorrect)
```python
if any(f.startswith(action) for f in self.config.filter.blocklist):
    return True
```

### Expected Code
```python
if any(action.startswith(f) for f in self.config.filter.blocklist):
    return True
```

### Analysis

The docstring for `blocklist` at L41-47 clearly states:
```python
blocklist: list[str] = [
    "vim",
    "vi",
    "emacs",
    ...
]
"""Block any command that starts with one of these"""
```

The intent is to block commands like `vim test.py` or `python -m venv env` because they **start with** a blocklisted string.

### Current Behavior

- `action = "vim test.py"` → **NOT blocked** (incorrect)

### Expected Behavior

- `action = "vim test.py"` → **blocked** because `"vim test.py".startswith("vim")` is True

### Impact

Interactive commands like `vim`, `emacs`, `nano`, etc. that should be blocked are currently allowed through, which could cause the agent to hang or behave unexpectedly when these commands are executed.

### Suggested Fix
```diff
- if any(f.startswith(action) for f in self.config.filter.blocklist):
+ if any(action.startswith(f) for f in self.config.filter.blocklist):
```

### Steps/commands/code to Reproduce

```bash
sweagent run \\
  --config {config_path} \\
  --agent.model.name={model_name} \\
  --agent.model.per_instance_cost_limit=240 \\
  --env.repo.path={repo_path} \\
  --output_dir={output_dir} \\
  --problem_statement.text="This is a software testing task. Generate test cases based on the provided prompts."
```

### Error message/results

...

### System Information

...

### Checklist

- [x] I'm running with the latest docker container/on the latest development version (i.e., I ran `git pull`))
- [x] I have enclosed code/log messages in triple backticks ([docs](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#quoting-code)) and clicked "Preview" to make sure it's displayed correctly.

thecaptain789:
## Description

The `should_block_action` method in `ToolHandler` had inverted logic when checking the blocklist. It was checking if blocklist items start with the action, instead of checking if the action starts with blocklist items.

## The Bug

```python
# Before (incorrect)
if any(f.startswith(action) for f in self.config.filter.blocklist):
    return True
```

This checks if `'vim'` starts with `'vim test.py'` → **False** (command not blocked)

```python
# After (correct)  
if any(action.startswith(f) for f in self.config.filter.blocklist):
    return True
```

This checks if `'vim test.py'` starts with `'vim'` → **True** (command blocked ✓)

## Impact

Interactive commands like `vim`, `emacs`, `nano`, etc. that should be blocked were being allowed through, potentially causing the agent to hang when these commands are executed.

## Related Issue

Fixes #1316

### DIFF
--- a/sweagent/tools/tools.py
+++ b/sweagent/tools/tools.py
@@ -355,7 +355,7 @@ class ToolHandler:
         action = action.strip()
         if not action:
             return False
-        if any(f.startswith(action) for f in self.config.filter.blocklist):
+        if any(action.startswith(f) for f in self.config.filter.blocklist):
             return True
         if action in self.config.filter.blocklist_standalone:
             return True


klieret:
Thank you!

PR #1343, opened 2026-02-17
feat: add private repository support for SWE-smith instances

mkaramuk:
## Summary

SWE-agent already supports cloning private repos via GITHUB_TOKEN in GithubRepoConfig. This PR addresses a different scenario: SWE-smith containers come with the repo pre-installed at /testbed, so there is no clone step. During environment reset, the container needs to fetch a specific bug-branch commit from the repo's private mirror — which requires SSH key injection into the reset commands.

- Adds SWESmithRepoConfig, a new repo type that handles SSH key injection and targeted fetch from private GitHub mirrors during environment reset
- Rewrites SWESmithInstances.get_instance_configs() to automatically detect private repos via the GitHub API and configure SSH-based git fetch when needed
- Adds _is_repo_private() and _find_and_encode_ssh_key() utilities to sweagent/utils/github.py
- Uses Pydantic SecretStr for SSH key storage, consistent with the existing api_key pattern in GenericAPIModelConfig
- Includes .env.example documenting the GITHUB_TOKEN and GITHUB_USER_SSH_KEY environment variables

## How it works

When loading SWE-smith task instances, each instance's repo field is checked against the GitHub API. If the repo is private (or returns 404), the loader:
1. Discovers an SSH key from GITHUB_USER_SSH_KEY env var or default ~/.ssh/ paths
2. Sets mirror_url to the SSH git URL of the private repo
3. During environment reset, SWESmithRepoConfig.get_reset_commands() injects the base64-decoded SSH key and fetches the bug branch from the mirror

Public repos continue to work as before with no SSH setup.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -6,10 +6,3 @@
 #   public_repo  - for public repos (read metadata, open PRs)
 #   repo         - for private repos (all of the above + private repo access)
 GITHUB_TOKEN=
-
-# Custom SSH key for private repo operations (optional)
-# If not set, the system looks for default keys in ~/.ssh/
-# (id_rsa, id_ecdsa, id_ecdsa_sk, id_ed25519, id_ed25519_sk)
-#
-# Set this only if your GitHub SSH key is at a non-standard path.
-# GITHUB_USER_SSH_KEY=/home/user/.ssh/id_ed25519_github
--- a/sweagent/environment/repo.py
+++ b/sweagent/environment/repo.py
@@ -6,7 +6,7 @@ from typing import Any, Literal, Protocol
 
 from git import InvalidGitRepositoryError
 from git import Repo as GitRepo
-from pydantic import BaseModel, ConfigDict, Field, SecretStr
+from pydantic import BaseModel, ConfigDict, Field
 from swerex.deployment.abstract import AbstractDeployment
 from swerex.runtime.abstract import Command, UploadRequest
 from typing_extensions import Self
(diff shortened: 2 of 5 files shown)


klieret:
But why wouldn't setting the `GITHUB_TOKEN` not also make `git fetch`/`git reset` work?

klieret:
Also pinging @john-b-yang since this is about SWE-smith

mkaramuk:
Hey @klieret, thanks for pointing out! Since we had implemented that SSH based auth in [SWE-smith](https://github.com/SWE-bench/SWE-smith/pull/220) (using token in SWE-smith caused baking it in the container image layers), thought that we can just apply the same approach but you are right. Using the token would be sufficient here since there is no container image building logic. We've updated the code so you can check it.

klieret:
Thanks, so I guess at this point the main difference with the other envs is that there's no cloning step for SWE-smith but you still want to fetch?

klieret:
Open to merge this, but note that most of our maintenance work has shifted over to https://github.com/swe-agent/mini-swe-agent/ (which I think would also be a great fit for smith), so swe-agent probably has a few issues (also with model support).
However, I'd actually be happy to support you fixing some of these. SWE-agent still has a relatively large userbase (though I think that 99% of usage would be better served with mini, but I understand that it might take some time to migrate).

klieret:
Sorry for the late merge!

Issue #1313, opened 2025-11-14
swe-bench Docker image uses Python 3.5, so that tool `edit_anthropic` can't run at all


stargazerZJ:
### Describe the bug

As has been discovered in #1179, some swebench testcases use Python 3.5, which is incompatible with `tree-sitter==0.21.3` required by edit_anthropic

After that has been fixed, I further discover that Python 3.5 is not compatible with the code of `edit_anthropic`. That tool just can't run at all, leaving the agent with an error message

### Steps/commands/code to Reproduce

```

#!/bin/bash

set -e

EVAL_SWEAGENT_VERSION="1.1.0"

INSTANCE_ID="django__django-7530"
OUTPUT_DIR="outputs/$INSTANCE_ID"
DATASET="princeton-nlp/SWE-bench_Verified"
SUBSET="verified"

CUSTOM_LLM_PROVIDER="litellm_proxy"
BASE_URL="http://127.0.0.1:8000/v1"
API_KEY="token-abc123"
MODEL="claude-3-7-sonnet-20250219"
TEMPERATURE="0.0"

# clean up output directory if it exists
rm -rf $OUTPUT_DIR

# run inference
/root/miniconda3/envs/swe-agent-v$EVAL_SWEAGENT_VERSION/bin/sweagent run-batch \
    --output_dir $OUTPUT_DIR \
    --instances.filter=$INSTANCE_ID \
    --instances.type=swe_bench \
    --instances.subset=$SUBSET \
    --instances.split=test \
    --instances.shuffle=False \
    --instances.evaluate=False \
    --instances.deployment.startup_timeout=600 \
    --agent.model.name=$CUSTOM_LLM_PROVIDER/$MODEL \
    --agent.model.api_base=$BASE_URL \
    --agent.model.api_key=$API_KEY \
    --agent.model.temperature=$TEMPERATURE \
    --agent.model.per_instance_call_limit="150" \
    --agent.model.retry.retries=30 \
    --agent.tools.parse_function.type="function_calling" \
    --config config/benchmarks/250225_anthropic_filemap_simple_review.yaml
```

### Error message/results

```
OBSERVATION:
  File "/root/tools/edit_anthropic/bin/str_replace_editor", line 26
    TRUNCATED_MESSAGE: str = "<response clipped><NOTE>To save on context only part of this file has been shown to you. You should retry this tool after you have searched inside the file with `grep -n` in order to find the line numbers of what you are looking for.</NOTE>"
                     ^
[...]

PR #1345, opened 2026-02-19
Fix Python 3.5 compatibility in edit_anthropic tool

paipeline:
Fixes issue #1313


klieret (changes requested):
Thanks!

Will merge if you fix the commented out lines (incorrect conversion of `{var=}` format)


PS: For almost all research i'd recommend https://github.com/swe-agent/mini-swe-agent/ over swe-agent

PR #1346, opened 2026-02-23
Add User-Agent header for LLM API calls

mikelambert:

Passes User-Agent: swe-agent/{version} via extra_headers to LiteLLM so providers can identify traffic from SWE-agent.

Changes:
* sweagent/agent/models.py
* tests/test_models.py

### DIFF
--- a/sweagent/agent/models.py
+++ b/sweagent/agent/models.py
@@ -25,7 +25,7 @@ from tenacity import (
     wait_random_exponential,
 )
 
-from sweagent import REPO_ROOT
+from sweagent import REPO_ROOT, __version__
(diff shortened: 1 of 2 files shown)


mikelambert:
Thanks, I see it's approved, but don't see a button to merge. Is this something you need to merge on your end?

klieret:
(was just waiting for tests to complete ;) )

klieret:
But btw, heavily recommend https://github.com/SWE-agent/mini-swe-agent/ for almost all workflows instead of SWE-agent, SWE-agent is now in maintenance only mode

PR #1399, opened 2026-04-28
Security: add confirmation gate for model-controlled bash commands

johnpippett:
Fixes a critical security finding where the agent executed every model-generated bash command immediately with zero confirmation gate.

- Added opt-in confirm_actions boolean to DefaultAgentConfig and ShellAgentConfig
- When confirm_actions=True, the agent prompts Execute? [Y/n] before running commands
- When confirm_actions=False (default, for SWE-bench backward compatibility), a warning is logged at init

57 tests pass; 2 xfailed (pre-existing).

### DIFF
--- a/sweagent/agent/agents.py
+++ b/sweagent/agent/agents.py
@@ -161,6 +161,13 @@ class DefaultAgentConfig(BaseModel):
     """
     action_sampler: ActionSamplerConfig | None = None
 
+    confirm_actions: bool = False
+    """Whether to prompt for human confirmation before executing each action.
+    Enabling this mitigates the risk of arbitrary shell execution from
+    model-controlled or prompt-injected commands. Defaults to False for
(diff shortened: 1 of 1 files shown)


arian-gogani:
confirmation gates for model-controlled bash are critical. the addition: a signed receipt for every gate decision.

when the gate allows or blocks a bash command, that decision should be captured in a tamper-evident receipt. the receipt carries: the command that was proposed, the gate's verdict (allow/deny), and the policy version. if a destructive command gets through and causes damage, the receipt chain shows exactly what was approved and by what policy.

we built this receipt format. Ed25519 signed, JCS canonical. the verdict field is specifically designed for allow/deny gate patterns.

github.com/arian-gogani/nobulex

klieret:
out of scope for this project. https://github.com/swe-agent/mini-swe-agent/ does have sth like this

PR #1418, opened 2026-05-30
docs: add Tuning Engines model route

cerebrixos:
Hi SWE-agent maintainers, we are an AI infrastructure team using SWE-agent with governed OpenAI-compatible endpoints. Since SWE-agent already routes through LiteLLM and supports api_base, I added a small Tuning Engines example to the existing model setup docs.

The intent is to keep SWE-agent in charge of the repository loop, tools, and parsers, while letting teams optionally route model traffic through Tuning Engines for tenant policy, traces, approvals, and usage visibility.

It would mean a lot to us if you are open to including this alongside the existing LiteLLM/OpenAI-compatible guidance.

Validation:
- git diff --check


klieret:
Hi thanks, but for now I'd keep things more concise.

PR #1425, opened 2026-06-18
harden(tools): block interactive commands in wrapped/compound forms

Jiangrong-W:
#### Reference Issues/PRs

None — found while reading the tool-filtering code.

**Summary.** This hardens `ToolHandler.should_block_action` so the
interactive-command blocklist sees the command line the way the shell will
actually run it. The blocklist is the reliability filter that stops the agent
from launching interactive / long-running programs (`vim`, `less`, `tail -f`, a
bare `python` REPL, `make`, ...) that would otherwise hang the run loop. Today
it only inspects the **first token of the whole action string** and normalizes
it with `action.strip()`, so a blocked program is still launched whenever the
model writes it in any non-bare shape.

**Scope (important).** This is **reliability hardening of an interactive-command
filter, not a security control.** The agent already has arbitrary shell
execution in its (Docker) environment by design — `handle_action` feeds
`step.action` straight into `env.communicate()` → `BashAction`. The blocklist
exists only to keep the run loop from stalling on interactive programs;
re-spelling a blocked command grants no extra capability. The change just makes
the filter robust to the obvious ways a command can be re-written so the agent
does not re-enter a program that hangs the run.

**Root cause.** The decision point and the execution point used different views
of the command string. The gate matched lexically against the first token only
(`startswith` over `blocklist`, exact match over `blocklist_standalone`), while
the shell runs the whole, fully-expanded command line. So path-qualified,
wrapper-prefixed, operator-wrapped and command-substitution forms all passed the
gate but ran normally:

- path-qualified: `/usr/bin/vim file`
- wrapper-prefixed: `env vim`, `sudo nano /etc/hosts`, `FOO=bar /usr/bin/vim`
- operator-wrapped / compound: `true; vim`, `echo x | less`, `cat f && less g`
- command substitution: `$(vim)`, `` `vim` ``

**Affected sites (decision point that under-normalized the command):**

- `sweagent/tools/tools.py:358` — `if any(action.startswith(f) for f in self.config.filter.blocklist):`
  (first-token-only lexical check; `action.strip()` only trims whitespace, so the
  wrapped/compound forms above are not recognized).

**The change (one place, `sweagent/tools/tools.py`).**

- `_iter_command_invocations(action)` splits a (possibly compound) command line
  into its individual invocations using a **quote-aware** tokenizer
  (`shlex.shlex(..., punctuation_chars=True)`) and also pulls out the inner
  commands of `$(...)` / back-tick substitutions. Quote-awareness means
  operators inside a quoted argument (`echo "a && vim"`) do **not** spuriously
  split the line; on unbalanced quotes it falls back to a naive split so obvious
  invocations still surface.
- `_normalize_invocation(segment)` strips leading `VAR=value` assignments and
  `env`/`sudo`/`nice`/`exec`/… wrapper programs (and their options) and reduces
  the executable to its **basename**, so `/usr/bin/vim` and `env vim` match the
  same rules a bare `vim` already does.
- `should_block_action` now checks the whole original action **first** (so
  multi-word blocklist entries like `tail -f` / `python -m venv` and the
  `block_unless_regex` rules behave exactly as before), **plus** each parsed
  invocation and its normalized form. The per-command rule logic itself is
  unchanged — it is just factored into `_is_blocked_invocation` and run against
  each candidate. One minimal change closes all of the wrapped/compound forms at
  once because they all reduce to a candidate the existing rules already match.

**Backwards compatibility.**

- No public API change: `should_block_action(action: str) -> bool` is identical;
  the new helpers are additive.
- `shlex` is in the standard library and is already imported elsewhere in the
  package.
- Existing allowed commands stay allowed — including `python script.py`,
  `radare2 -c …` (the `block_unless_regex` allow rule), and blocklisted names
  used as **quoted arguments** (`echo "tail -f"`, `git commit -m 'x; vim'`).
  Existing blocked bare forms stay blocked. Pre-existing over-blocks from the
  original `startswith` rule (e.g. `vimtutor`) are unchanged.

**Tests.** Adds `tests/test_should_block_action.py`:

- `test_blocked_invocations` (parametrized) — bare, path-qualified,
  wrapper-prefixed, operator-wrapped / compound, and command-substitution forms
  of blocklisted commands must be blocked. These cases **fail on `main`** and
  **pass with this change** (on BASE the file reports `15 failed, 16 passed`).
- `test_allowed_invocations` (parametrized) — ordinary commands, blocklisted
  names as quoted arguments, non-interactive subcommands, and the `radare2 -c`
  allow rule must NOT be blocked.
- `test_empty_action_not_blocked`.

Files modified: `sweagent/tools/tools.py` (the filter) and
`tests/test_should_block_action.py` (new test).

Verification (local, on this branch): the new test file passes (`31 passed`);
the broader `tests/test_agent.py` / `tests/test_parsing.py` /
`tests/test_tools_command_parsing.py` suites continue to pass (`57 passed, 2
xfailed`, the 2 xfails pre-existing and unrelated); `ruff check` and `ruff
format --check` are clean on the changed files.

### DIFF
--- a/sweagent/tools/tools.py
+++ b/sweagent/tools/tools.py
@@ -9,6 +9,7 @@ import asyncio
 import json
 import os
 import re
+import shlex
 from functools import cached_property
 from pathlib import Path
 from typing import Any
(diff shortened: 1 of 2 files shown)


klieret:
Thanks, but if models still mess up despite the blocklist, that's on them. It's also probably not even needed anymore, https://github.com/SWE-agent/mini-swe-agent/ runs without it.

Issue #1159, opened 2025-05-24
Exiting due to unknown error: 'NoneType' object is not iterable


SqooAI:
### Describe the bug

just crash with :Exit due to unknown error: 'NoneType' object is not iterable:

![Image](https://github.com/user-attachments/assets/b1e64919-9fdb-46b9-9ca8-4ff14b61531f)

### Steps/commands/code to Reproduce

sweagent run --config=config/github.yaml --env.repo.github_url=https://github.com/Xpykerz/PassWord-Factory --problem_statement.github_url=https://github.com/Xpykerz/PassWord-Factory/issues/4 --actions.open_pr=true


### Error message/results

            Keep your changes minimal and focused.                                                                                                                                                                                          
🤠 INFO     ========================= STEP 1 =========================                                                                                                                                                                      
🤖 DEBUG    n_cache_control: 0                                                                                                                                                                                                              
🔧 DEBUG    Choosing API key 0 for thread MainThread (idx 0)                                                                                                                                                                                
🤖 INFO     Response: ModelResponse(id='chatcmpl-BaoA2lzTlSMtKKRxbp6GB17uA1jTc', created=1748112326, model='gpt-4o-2024-11-20', object='chat.completion', system_fingerprint='fp_ee1d74bde0', choices=[Choices(finish_reason='tool_calls',  
            index=0, message=Message(content=None, role='assistant', tool_calls=[ChatCompletionMessageToolCall(function=Function(arguments='{"command":"view","path":"/Xpykerz__PassWord-Factory","view_range":null}',                      
            name='str_replace_editor'), id='call_akj1SN8KqdPVeyQWWcdGxZ9J', type='function')], function_call=None, provider_specific_fields={'refusal': None, 'padding': 'abcdefghijklmno'}))], usage=Usage(completion_tokens=34,           
            prompt_tokens=864, total_tokens=898, completion_tokens_details=CompletionTokensDetailsWrapper(accepted_prediction_tokens=0, audio_tokens=None, reasoning_tokens=None, rejected_prediction_tokens=0, text_tokens=None),          
            prompt_tokens_details=PromptTokensDetailsWrapper(audio_tokens=None, cached_tokens=0, text_tokens=None, image_tokens=None)), service_tier=None, prompt_filter_results=[{'content_filter_results': {'hate': {'filtered': False,   
[...]

PR #1430, opened 2026-06-24
Fix NameError when tool call arguments are already a dict

Chessing234:
#### Reference Issues/PRs

Fixes a crash path related to #1159.

`FunctionCallingParser._parse_tool_call` only assigns `values` when `arguments` is a JSON string. Some providers (via LiteLLM) pass an already-parsed `dict`, so the next line hits `NameError: name 'values' is not defined` and kills the run instead of returning a format error to the model.

Add the missing `else` branch to use the dict directly.

Made with [Cursor](https://cursor.com)


klieret (approved):
thx!

Issue #1310, opened 2025-10-13
Add Trajectory Hook for Remote Environments


ColeMurray:
### Describe the feature

When running in remote environments, the patch is obtained as part of the SaveApplyPatchHook and saved to the launching environment. There is currently no similar hook for obtaining the trajectory.

To improve this experience for remote environments, we should add a SaveTrajectoryHook that would also save the traj in the launching env

### Potential Solutions

```python
"""Hook for saving agent trajectories to the local filesystem."""

import json
from pathlib import Path

from sweagent.agent.problem_statement import ProblemStatement, ProblemStatementConfig
from sweagent.environment.swe_env import SWEEnv
from sweagent.run.hooks.abstract import RunHook
from sweagent.types import AgentRunResult
from sweagent.utils.log import get_logger


class SaveTrajectoryHook(RunHook):
    """Saves agent trajectories to the local filesystem.

    Writes trajectory data to a .traj JSON file in the output directory,
    matching the internal .traj file format for compatibility with existing tooling.
    """

    def __init__(self, pretty_print: bool = True):
        """Initialize the SaveTrajectoryHook.

        Args:
            pretty_print: If True, formats JSON with indentation for readability.
        """
        self.logger = get_logger("swea-save_trajectory", emoji="📝")
        self._pretty_print = pretty_print
        self._output_dir: Path | None = None
        self._problem_statement: ProblemStatement | ProblemStatementConfig | None = None

    def on_init(self, *, run):
        """Called when hook is initialized."""
        self._output_dir = Path(run.output_dir)
        self.logger.info(f"Trajectory hook initialized with output_dir: {self._output_dir}")

    def on_instance_start(
        self,
        *,
        index: int,
        env: SWEEnv,
        problem_statement: ProblemStatement | ProblemStatementConfig,
    ):
        """Called at the beginning of each instance."""
        self._problem_statement = problem_statement
        self.logger.debug(f"Starting instance {index}: {problem_statement.id}")

    def on_instance_completed(self, *, result: AgentRunResult):
        """Called when an instance is completed. Saves the trajectory to a .traj file."""
        if self._problem_statement is None:
            self.logger.warning("No problem statement available, skipping trajectory save")
            return

        if self._output_dir is None:
            self.logger.error("Output directory not set, cannot save trajectory")
            return

        instance_id = self._problem_statement.id
        trajectory_path = self._save_trajectory(instance_id, result)

        if trajectory_path:
[...]

Issue #1284, opened 2025-08-21
SaveApplyPatchHook saves patch to wrong instance directory in batch setting


18jeffreyma:
### Describe the bug

https://github.com/SWE-agent/SWE-agent/blob/b2ca9c3262157382d0b1dfe8d5b0a224edbc7b83/sweagent/run/run_batch.py#L180

When running SWE-Bench evals, I noticed that the SaveApplyPatch hook is saving the wrong instance patch into another instance's log folder. This seems to be a bug due to the hook not actually having unique state per worker (relying on multi-worker on_start and on_end to manage state): the bug should show up with high number of workers where one instance starts, another starts, and one instance finishes before the other finishes.

### Steps/commands/code to Reproduce

(see above)

### Error message/results

No explicit error message, but .patch file is saved in wrong place

### System Information

Linux c7.xlarge VM

### Checklist

- [x] I'm running with the latest docker container/on the latest development version (i.e., I ran `git pull`))
- [x] I have copied the full command/code that I ran (as text, not as screenshot!)
- [x] If applicable: I have copied the **full** log file/error message that was the result (as text, not as screenshot!)
- [x] I have enclosed code/log messages in triple backticks ([docs](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#quoting-code)) and clicked "Preview" to make sure it's displayed correctly.

18jeffreyma:
more specifically, i believe this happens due to using a thread pool executor

https://github.com/SWE-agent/SWE-agent/blob/b2ca9c3262157382d0b1dfe8d5b0a224edbc7b83/sweagent/run/run_batch.py#L276

the hooks defined at the top of RunBatch are in the same shared memory space (and so the same SaveApplyPatchHook object is being modified for all worker threads in pool)

PR #1436, opened 2026-06-29
feat: add SaveTrajectoryHook for remote environments

anxkhn:
#### Reference Issues/PRs

Refs #1310

In remote/cloud deployments the agent writes its `.traj` inside the container's `output_dir`, and only the patch comes back to the launching environment via `SaveApplyPatchHook`. There is no equivalent hook to persist the trajectory. #1310 sketches a `SaveTrajectoryHook` to close that gap.

This adds `SaveTrajectoryHook(RunHook)` (`sweagent/run/hooks/save_trajectory.py`), modeled on `SaveApplyPatchHook`:

- Runs in the launching environment. `on_init` captures `run.output_dir`; `on_instance_completed` writes `output_dir/<id>/<id>.traj` from the `AgentRunResult` it receives, so no remote/container access is needed.
- Per-instance state is kept in `threading.local()`, matching `SaveApplyPatchHook`, so concurrent run-batch workers do not clobber one another (cf. #1284).
- `pretty_print` toggle for the JSON output.

Two things I want to be upfront about:

1. **It is not auto-registered.** Local runs already write a richer `.traj` via `Agent.save_trajectory`, so wiring this on by default would double-write / overwrite that file. I left it unregistered and net-additive. Happy to wire it in wherever you prefer: an opt-in config flag, or always-on only for the remote deployment paths. What is your preference?
2. **The document is a subset**, `{"trajectory", "info"}`, which is what `AgentRunResult` exposes. The internal `.traj` also carries `history`, env and replay config, so this is not byte-identical to it. Documented in the docstring.

Tests added to `tests/test_run_hooks.py`: per-instance path, no-op when no problem statement, and a concurrency regression guard mirroring #1284. Targeted suite (`tests/test_run_hooks.py`) is green and ruff check/format are clean.

#### Test plan

- `uv run --no-sync pytest tests/test_run_hooks.py -q` -> 12 passed (3 new)
- `uvx ruff@0.15.17 check` + `format --check` on changed files -> clean

### DIFF
--- a/sweagent/run/hooks/save_trajectory.py
+++ b/sweagent/run/hooks/save_trajectory.py
@@ -1,0 +1,69 @@
+"""Hook for saving agent trajectories to the launching environment's filesystem."""
+
+import json
+import threading
+from pathlib import Path
+
+from sweagent.agent.problem_statement import ProblemStatementConfig
+from sweagent.environment.swe_env import SWEEnv
+from sweagent.run.hooks.abstract import RunHook
+from sweagent.types import AgentRunResult
+from sweagent.utils.log import get_logger
+
+
+class SaveTrajectoryHook(RunHook):
+    """Saves agent trajectories to the launching environment's filesystem.
+
+    For remote/cloud deployments the agent writes its ``.traj`` inside the
+    container's ``output_dir``; only the patch is brought back to the launching
+    environment via :class:`~sweagent.run.hooks.apply_patch.SaveApplyPatchHook`.
+    This hook is the trajectory equivalent: it runs in the launching environment
+    and persists the trajectory from the :class:`~sweagent.types.AgentRunResult`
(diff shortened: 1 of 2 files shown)


anxkhn:
rebased onto current main and resolved the test-file conflict while preserving both sets of coverage. the focused hook tests and ruff checks pass. ready for re-review.

anxkhn:
covered the remaining codecov gap: tests now exercise `on_init` and `pretty_print=False`. local `tests/test_run_hooks.py` is 18 passed with 100% coverage on `save_trajectory.py`.

anxkhn:
covered the remaining codecov gap: tests now exercise `on_init` and `pretty_print=False`. local `tests/test_run_hooks.py` is 18 passed with 100% coverage on `save_trajectory.py`.

Issue #1182, opened 2025-05-31
Function calling command parsing: tuple of ints gets turned into string


klieret:
Resulting in things like this:

```
THOUGHT:


ACTION:
str_replace_editor view /testbed/django/db/migrations/serializer.py  --view_range [ 1 ,   5 0 ]

OBSERVATION:
usage: str_replace_editor [-h] [--file_text FILE_TEXT]
                          [--view_range VIEW_RANGE VIEW_RANGE]
                          [--old_str OLD_STR] [--new_str NEW_STR]
                          [--insert_line INSERT_LINE]
                          command path
str_replace_editor: error: argument --view_range: invalid int value: '['
```

despite `view_range` being passed correctly:

```
2025-05-26 22:45:45,550 - INFO - swea-lm-astropy__astropy-12907 - Response: ModelResponse(id='chatcmpl-9b470b7e-fd81-4940-822d-86bfc89a68a6', created=1748299545, model='claude-sonnet-4-20250514', object='chat.completion', system_fingerprint=None, choices=[Choices(f
     inish_reason='tool_calls', index=0, message=Message(content="Great! I've reproduced the issue. The problem is that when we have a nested compound model like `m.Pix2Sky_TAN() & (m.Linear1D(10) & m.Linear1D(5))`, the separability matrix is not computed correctly. The
      last two rows should be `[False, False, True, False]` and `[False, False, False, True]` but we get `[False, False, True, True]` and `[False, False, True, True]`.\n\nLet's examine the core.py file to understand how CompoundModel works:", role='assistant', tool_call
     s=[ChatCompletionMessageToolCall(index=1, function=Function(arguments='{"command": "view", "path": "/testbed/astropy/modeling/core.py", "view_range": [1, 100]}', name='str_replace_editor'), id='toolu_01RSApM59ukD75uiqCFMDerD', type='function')], function_call=None,
      provider_specific_fields={'citations': []}, citations=[]))], usage=Usage(completion_tokens=242, prompt_tokens=12641, total_tokens=12883, completion_tokens_details=None, prompt_tokens_details=PromptTokensDetailsWrapper(audio_tokens=None, cached_tokens=11689, text_t
     okens=None, image_tokens=None), cache_creation_input_tokens=945, cache_read_input_tokens=11689))
```

justinchiu-cohere:
[...]

PR #1449, opened 2026-07-07
fix: reject malformed view_range in XMLFunctionCallingParser

anxkhn:
#### Reference Issues/PRs

None. This is a self-contained fix; I did not find an existing issue or PR for
it. It is a sibling of the `FunctionCallingParser` argument-handling work in
#1417 / #1182, but it is a different parser and a different code path, so it does
not overlap with that PR.

`XMLFunctionCallingParser` validates the `view_range` parameter with an
`re.match` that is only anchored at the start of the string:

```python
# sweagent/tools/parsing.py
v = params_dict["view_range"]
if isinstance(v, str):
    if not re.match(r"\[\d+,\s*\d+\]", v):
        msg = f"view_range must be in the format [<start>, <end>], got {v}."
        raise FormatError(msg)
    params_dict["view_range"] = json.loads(v)
```

Because `re.match` is unanchored at the end, a value that starts with a valid
`[<int>, <int>]` prefix but has trailing junk, for example `"[1,2]extra"` or
`"[1, 2] and more"`, passes the format check. It is then handed to
`json.loads`, which is not valid JSON for a two-element list and raises
`json.JSONDecodeError`.

That exception is the problem. `JSONDecodeError` is a subclass of `ValueError`,
not `FormatError`. The parser is supposed to raise `FormatError` here so the
agent catches it and re-queries the model to fix its output format. The uncaught
`JSONDecodeError` instead falls through to the broad error handler, which aborts
the whole run for that instance via the auto-submission/exit path. So a model
that emits a `view_range` with a valid numeric prefix plus stray trailing
characters makes SWE-agent give up on the instance instead of nudging the model
to correct the format.

The fix is one line: use `re.fullmatch` so the whole string must conform. Any
malformed `view_range` now takes the existing `FormatError` branch (with the
already-present, helpful message) instead of reaching `json.loads`.

Changes:
- `sweagent/tools/parsing.py`: `re.match` -> `re.fullmatch` in the
  `XMLFunctionCallingParser` `view_range` validation.
- `tests/test_parsing.py`: add `test_xml_function_calling_parser_view_range`,
  which drives the parser end-to-end and asserts a well-formed value
  (`[1, 20]`) parses into the action, while `[1,2]extra` and `[1, 2] and more`
  raise `FormatError` (previously they raised an uncaught `JSONDecodeError`).

One small, intended behavior change worth calling out: `re.fullmatch` also
rejects a trailing space, for example `"[1,2] "`, which `json.loads` used to
accept (JSON ignores trailing whitespace). This now yields a clean `FormatError`
re-query rather than silent acceptance, which matches the documented format
`[<start>, <end>]`.

Tested with `pytest tests/test_parsing.py` (9 passed); `ruff check` and
`ruff format` are clean on the changed files.

### DIFF
--- a/sweagent/tools/parsing.py
+++ b/sweagent/tools/parsing.py
@@ -285,7 +285,7 @@ class XMLFunctionCallingParser(AbstractParseFunction, BaseModel):
             # Check that value is format as [x, y]
             v = params_dict["view_range"]
             if isinstance(v, str):
-                if not re.match(r"\[\d+,\s*\d+\]", v):
+                if not re.fullmatch(r"\[\d+,\s*\d+\]", v):
                     msg = f"view_range must be in the format [<start>, <end>], got {v}."
                     raise FormatError(msg)
                 params_dict["view_range"] = json.loads(v)
--- a/tests/test_parsing.py
+++ b/tests/test_parsing.py
@@ -4,7 +4,7 @@ import pytest
 from jinja2 import Template
 
 from sweagent.exceptions import FormatError, FunctionCallingFormatError
-from sweagent.tools.commands import Command
+from sweagent.tools.commands import Argument, Command
 from sweagent.tools.parsing import (
     ActionParser,
     EditFormat,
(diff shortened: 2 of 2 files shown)


anxkhn:
@klieret this parser validation fix remains clean and green. could you review it when you have a chance?

PR #1453, opened 2026-07-10
fix: fully validate command argument names

anxkhn:
#### Reference Issues/PRs

No existing issue. This is a small correctness fix found while reviewing command argument validation.

`Command.validate_arguments` used `re.match` with `ARGUMENT_NAME_PATTERN`, which only validates a prefix of the argument name. Names such as `"arg name"`, `"a/b"`, and `"arg$name"` therefore passed validation and produced malformed invocation formats that failed later with less useful errors.

This changes the check to `re.fullmatch` so the entire argument name must conform to the existing pattern. It also extends `test_argument_name_patterns` with names that have a valid prefix followed by invalid characters. Existing valid names, including dashed names such as `replace-all`, remain accepted.

### DIFF
--- a/sweagent/tools/commands.py
+++ b/sweagent/tools/commands.py
@@ -196,7 +196,7 @@ class Command(BaseModel):
             msg = f"Command '{self.name}': Duplicate argument names: {duplicates}"
             raise ValueError(msg)
         for arg in self.arguments:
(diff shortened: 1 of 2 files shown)


anxkhn:

Issue #889, opened 2025-02-05
Multiple toolcalls and DeepSeek-R1


Martovark:
### Describe the issue

Hello, everyone!
Faced a problem on running deepseek R1(distelled Llama-8b and Qwen-14b), namely, the model almost always calls multiple functions in a single response.  Deepseek team recommends using their models as follows:

<img width="819" alt="Image" src="https://github.com/user-attachments/assets/3e1b16df-68a4-43f4-a386-f9b1a0a3e3fd" />

- In the config (and in the demonstration) I merged system_template + instance_template -> instance_template
- Varying the temperature from 0.5 to 0.7.

But it doesn't help much. It also seems useful to remove the tokens and text between `<think>, </think>`, when adding an action to the history.
Has anyone had any luck getting the trajectories to run on the new R1 models?

klieret:
A few things

1. Maybe a demonstration could help?
2. You could write a [`HistoryProcessor`](https://swe-agent.com/latest/reference/history_processor_config/)  to modify actions when they're added to the history
3. You can write a custom parser to only take the first action
4. Could you use stop sequences to stop the generation of the LM, for example after it hits `</action>` or something like that?

Cross linking #891

ofirpress:
Llama-8b / Qwen-14b sounds too small. I don't think we should spend much time solving issues with such small models at this point. I guess they might improve in the future, but for now I'd just use slightly bigger models...

klieret:
Closing this for lack of follow up

PR #1456, opened 2026-07-12
fix: forward configured stop sequences to litellm.completion

anxkhn:
#### Reference Issues/PRs

See also #889 (users asking to constrain output for DeepSeek-R1 style models).
No open issue tracks this specific gap; found while reading the model config.

`GenericAPIModelConfig` exposes a `stop` field, documented as "Custom stop
sequences" and settable from any model config:

```python
# sweagent/agent/models.py
stop: list[str] = []
"""Custom stop sequences"""
```

`stop` is a first-class, OpenAI-standard parameter of `litellm.completion`
(up to 4 sequences at which the provider stops generating). However,
`LiteLLMModel._single_query` builds the completion call without it. It passes
`model`, `messages`, `temperature`, `top_p`, `api_version`, `api_key`,
`fallbacks`, `**completion_kwargs`, `**extra_args`, and `n`, but never `stop`:

```python
response = litellm.completion(
    model=self.config.name,
    messages=messages,
    temperature=...,
    top_p=self.config.top_p,
    api_version=self.config.api_version,
    api_key=self.config.choose_api_key(),
    fallbacks=self.config.fallbacks,
    **completion_kwargs,
    **extra_args,
    n=n,
)
```

`self.config.stop` is read nowhere else in the module, so any stop sequences a
user configures are silently dropped and never reach the provider. That is
surprising for open and local models that rely on stop sequences to terminate
their output.

This forwards `config.stop` through the existing `extra_args` block, and only
when it is non-empty, mirroring exactly how the adjacent optional `api_base`
argument is handled:

```python
if self.config.api_base:
    # Not assigned a default value in litellm, so only pass this if it's set
    extra_args["api_base"] = self.config.api_base
if self.config.stop:
    # Only pass custom stop sequences if configured, so the default behavior is unchanged
    extra_args["stop"] = self.config.stop
```

Because the default is an empty list (falsy), nothing is sent unless the user
opts in, so existing behavior is unchanged.

Files changed:
- `sweagent/agent/models.py`: forward `config.stop` to the completion call when set (+3 lines).
- `tests/test_models.py`: two regression tests: `stop` is forwarded when configured, and omitted when not.

Testing: `pytest tests/test_models.py` passes (6 passed), no Docker or LLM key
needed. Reverting the 3-line fix makes `test_stop_sequences_forwarded` fail
(the forwarded `stop` is `None`), confirming the test catches the regression.
`ruff check` and `ruff format --check` are clean on both files.

### DIFF
--- a/sweagent/agent/models.py
+++ b/sweagent/agent/models.py
@@ -705,6 +705,9 @@ class LiteLLMModel(AbstractModel):
         if self.config.api_base:
             # Not assigned a default value in litellm, so only pass this if it's set
             extra_args["api_base"] = self.config.api_base
+        if self.config.stop:
+            # Only pass custom stop sequences if configured, so the default behavior is unchanged
+            extra_args["stop"] = self.config.stop
         if self.tools.use_function_calling:
             extra_args["tools"] = self.tools.tools
         # We need to always set max_tokens for anthropic models
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -104,3 +104,43 @@ def test_user_agent_header_with_other_extra_headers():
         extra_headers = call_kwargs.kwargs.get("extra_headers", {})
         assert extra_headers["User-Agent"] == f"swe-agent/{__version__}"
         assert extra_headers["X-Custom"] == "value"
+
+
+def test_stop_sequences_forwarded():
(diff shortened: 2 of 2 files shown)


anxkhn:

PR #1458, opened 2026-07-14
fix: map multimodal subset to sb-cli's swe-bench-m

anxkhn:
#### Reference Issues/PRs

No existing issue. This is a follow-up to #1223 (SWE-agent Multimodal), which
added the `multimodal` subset to `SWEBenchInstances` but did not update the
evaluation hook's subset mapping.

`SweBenchEvaluate` builds an `sb-cli submit` command whenever a batch run is
started with `--evaluate=True`. It translates the run's `subset` to the subset
identifier that sb-cli expects via `_SUBSET_MAP`
(`sweagent/run/hooks/swe_bench_evaluate.py`):

```python
_SUBSET_MAP = {"lite": "swe-bench_lite", "verified": "swe-bench_verified", "multimodal": "swe-bench_multimodal"}
```

Two problems:

1. **`multimodal` maps to a subset sb-cli rejects.** sb-cli's `Subset` enum only
   accepts `swe-bench_lite`, `swe-bench_verified` and `swe-bench-m`, and its
   `submit` command declares the subset argument with that enum type, so any
   other string is rejected at the CLI boundary. The multimodal dataset is
   `swe-bench-m`, not `swe-bench_multimodal`. As a result, running a multimodal
   batch with `--evaluate=True` always fails when the hook shells out to sb-cli.

2. **`full` and `multilingual` raise a bare `KeyError`.** Both are valid
   `SWEBenchInstances.subset` values for loading instances, but neither is a key
   in `_SUBSET_MAP`, so `_get_sb_call` fails with an unhelpful
   `KeyError: 'full'` / `KeyError: 'multilingual'` instead of a clear message.

Both symptoms share one root cause: `_SUBSET_MAP` drifted out of sync with
sb-cli's accepted subsets when the multimodal subset was introduced in #1223.

**Changes** (`sweagent/run/hooks/swe_bench_evaluate.py`):

- Map `"multimodal"` to `"swe-bench-m"` so multimodal evaluation submits a
  subset sb-cli accepts.
- Guard `_get_sb_call`: if the subset has no sb-cli equivalent (`full`,
  `multilingual`), raise a descriptive `ValueError` naming the supported
  subsets instead of a bare `KeyError`. This mirrors the existing "Unsupported
  subset" `ValueError` in `batch_instances.py`.

**Tests** (`tests/test_run_hooks.py`): parametrized regression tests that assert
the subset passed to sb-cli for `lite`/`verified`/`multimodal`, and that
`full`/`multilingual` raise `ValueError`. Verified failing before the fix and
passing after.

```
pytest tests/test_run_hooks.py -k swe_bench_evaluate   # 5 passed
```

### DIFF
--- a/sweagent/run/hooks/swe_bench_evaluate.py
+++ b/sweagent/run/hooks/swe_bench_evaluate.py
@@ -17,7 +17,10 @@ from sweagent.utils.log import get_logger
 
 
 class SweBenchEvaluate(RunHook):
-    _SUBSET_MAP = {"lite": "swe-bench_lite", "verified": "swe-bench_verified", "multimodal": "swe-bench_multimodal"}
+    # Maps SWEBenchInstances.subset values to the subset identifiers accepted by
+    # sb-cli. sb-cli only supports these three; "full" and "multilingual" have no
+    # sb-cli equivalent and therefore cannot be evaluated this way.
+    _SUBSET_MAP = {"lite": "swe-bench_lite", "verified": "swe-bench_verified", "multimodal": "swe-bench-m"}
 
     def __init__(self, output_dir: Path, subset: str, split: str, continuous_submission_every: int = 0) -> None:
         super().__init__()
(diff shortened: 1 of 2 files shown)


klieret (approved):

klieret:
test failures seem unrelated

Issue #981, opened 2025-02-26
Race condition: get_logger and modifying _ADDITIONAL_HANDLERS


klieret:
This is pretty rare and mostly when `KeyboardInterrupt`ing the agent

```python
🏃 ERROR    [swea-run] Traceback (most recent call last):
              File "/home/klieret/SWE-agent/sweagent/run/run_batch.py", line 300, in run_instance
                result = self._run_instance(instance)
              File "/home/klieret/SWE-agent/sweagent/run/run_batch.py", line 326, in _run_instance
                agent = get_agent_from_config(self.agent_config)
              File "/home/klieret/SWE-agent/sweagent/agent/agents.py", line 206, in get_agent_from_config
                return DefaultAgent.from_config(config)
                       ~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^
              File "/home/klieret/SWE-agent/sweagent/agent/agents.py", line 458, in from_config
                model = get_model(config.model, config.tools)
              File "/home/klieret/SWE-agent/sweagent/agent/models.py", line 781, in get_model
                return LiteLLMModel(args, tools)
              File "/home/klieret/SWE-agent/sweagent/agent/models.py", line 554, in __init__
                self.logger = get_logger("swea-lm", emoji="🤖")
                              ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
              File "/home/klieret/SWE-agent/sweagent/utils/log.py", line 79, in get_logger
                for handler in _ADDITIONAL_HANDLERS.values():
                               ~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
            RuntimeError: dictionary changed size during iteration
```

PR #1459, opened 2026-07-14
fix(log): guard shared logger globals writers with _LOG_LOCK

anxkhn:
#### Reference Issues/PRs

Recurrence of #981 (`Race condition: get_logger and modifying _ADDITIONAL_HANDLERS`).
That issue was closed by #993, but the fix there was incomplete (see below), so the
same crash can still happen. Not using a `Fixes` keyword because #981 is already
closed.

**The bug.** Under `run-batch` with `num_workers > 1`, `get_logger` can still crash with

```
RuntimeError: dictionary changed size during iteration
```

(or the `set` variant), which is the exact failure reported in #981.

`sweagent/utils/log.py` keeps two module-level globals that are shared across threads:
`_SET_UP_LOGGERS` (a `set`) and `_ADDITIONAL_HANDLERS` (a `dict`). `_LOG_LOCK` exists
to serialize access to them. #993 (and the earlier handler-locking commit) wrapped the
**readers** that iterate these containers in `with _LOG_LOCK:`, but left the **writers**
outside the lock:

- `get_logger` iterated `_ADDITIONAL_HANDLERS` under the lock, but called
  `_SET_UP_LOGGERS.add(name)` *outside* it.
- `add_file_handler` iterated `_SET_UP_LOGGERS` under the lock, but did
  `_ADDITIONAL_HANDLERS[id_] = handler` *outside* it.
- `remove_file_handler` iterated `_SET_UP_LOGGERS` under the lock, but did
  `_ADDITIONAL_HANDLERS.pop(id_)` *outside* it.

Holding the lock only while you read a container gives no protection when another
thread mutates that same container without holding the lock. So thread A iterating
`_ADDITIONAL_HANDLERS.values()` under the lock in `get_logger` can still be interrupted
by thread B doing an unlocked `_ADDITIONAL_HANDLERS[...] = ...` in `add_file_handler`,
and Python raises `dictionary changed size during iteration` mid-iteration. The
symmetric case holds for `_SET_UP_LOGGERS` (`get_logger`'s unlocked `.add` versus the
locked `for name in _SET_UP_LOGGERS` in `add_file_handler`/`remove_file_handler`). The
comments already said `Lock because other thread might be modifying the
_SET_UP_LOGGERS set`, but the code that actually modifies it did not take the lock, so
the invariant was not enforced.

This is reachable in normal use: `run_batch` runs `run_instance` for each instance in a
`ThreadPoolExecutor(max_workers=num_workers)`, and those worker threads both create
loggers (`get_logger`) and add/remove per-run file handlers concurrently.

**The fix.** Move each of the three writers inside the existing `with _LOG_LOCK:` block
so every access to `_SET_UP_LOGGERS` and `_ADDITIONAL_HANDLERS` is serialized:

- `get_logger`: `_SET_UP_LOGGERS.add(name)` now runs under the lock.
- `add_file_handler`: `_ADDITIONAL_HANDLERS[id_] = handler` now runs under the lock,
  atomically with adding the handler to the already-set-up loggers. The purely local
  setup (`handler.my_filter = filter` and the `id_` default) is computed before the lock
  so the critical section stays small; a side benefit is that `my_filter` is now always
  set before the handler becomes visible in `_ADDITIONAL_HANDLERS`, so `get_logger` can
  no longer observe a registered handler that is missing its `my_filter` attribute.
- `remove_file_handler`: `_ADDITIONAL_HANDLERS.pop(id_)` now runs under the lock.

`_LOG_LOCK` is a plain (non-reentrant) `threading.Lock`, and nothing inside these blocks
re-acquires it, so there is no deadlock. The lock/`logging` ordering is unchanged: the
readers already called `logging.getLogger(...)`/`logger.addHandler(...)` under
`_LOG_LOCK`. There is no functional change beyond the added serialization.

**Files changed**

- `sweagent/utils/log.py`: moved the three writers inside `_LOG_LOCK`; updated the two
  comments to note the lock also guards `_ADDITIONAL_HANDLERS`.
- `tests/test_log.py` (new): a regression test that hammers `get_logger` against
  `add_file_handler`/`remove_file_handler` from 12 threads and fails if any thread
  raises. To make the race reproducible rather than luck-based it pre-populates the
  globals (so the locked reader loops iterate large containers, widening the window) and
  lowers `sys.setswitchinterval`; it also temporarily detaches pytest's root-logger
  capture handlers so `get_logger` does not short-circuit at `hasHandlers()` and actually
  reaches the racy code path. All global/state changes (switch interval, root handlers,
  and every entry it added to the globals) are restored in `finally`. On `main` this test
  fails with the `RuntimeError: ... changed size during iteration`; with the fix it
  passes.

**How I verified**

- `pytest tests/test_log.py` passes with the fix and fails on unmodified `main`
  (reproducing both the `set` and the `dict` "changed size during iteration" errors).
- No regressions in the Docker-free unit subset (`tests/test_log.py`,
  `tests/test_models.py`, `tests/test_utils.py`, `tests/test_parsing.py`, and the
  concurrent-workers run-hook test).
- `ruff check` and `ruff format --check` are clean on both files.

### DIFF
--- a/sweagent/utils/log.py
+++ b/sweagent/utils/log.py
@@ -75,8 +75,8 @@ def get_logger(name: str, *, emoji: str = "") -> logging.Logger:
     logger.setLevel(logging.TRACE)  # type: ignore
     logger.addHandler(handler)
     logger.propagate = False
-    _SET_UP_LOGGERS.add(name)
     with _LOG_LOCK:
+        _SET_UP_LOGGERS.add(name)
         for handler in _ADDITIONAL_HANDLERS.values():
             my_filter = getattr(handler, "my_filter", None)
             if my_filter is None:
@@ -114,8 +114,12 @@ def add_file_handler(
     formatter = logging.Formatter("%(asctime)s - %(levelname)s - %(name)s - %(message)s")
     handler.setFormatter(formatter)
     handler.setLevel(_interpret_level(level))
+    handler.my_filter = filter  # type: ignore
+    if not id_:
+        id_ = str(uuid.uuid4())
     with _LOG_LOCK:
         # Lock because other thread might be modifying the _SET_UP_LOGGERS set
+        # or iterating _ADDITIONAL_HANDLERS.
         for name in _SET_UP_LOGGERS:
             if filter is not None:
                 if isinstance(filter, str) and filter not in name:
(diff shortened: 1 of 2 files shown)


anxkhn:

PR #1461, opened 2026-07-15
fix(replay): preserve environment startup configuration

anxkhn:
#### Reference Issues/PRs

None. This is a standalone replay correctness fix found while reviewing the trajectory replay path.

`RunReplay` parses the environment configuration recorded in a trajectory, but `_get_env()` previously reconstructed `SWEEnv` with `post_startup_commands=[]` and left the startup timeout and environment name at their defaults. As a result, replay skipped any recorded environment setup commands and could execute against a different environment than the original run, causing failures or divergent results.

This forwards `post_startup_commands`, `post_startup_command_timeout`, and `name` from the recorded environment configuration while retaining the deployment selected for the replay. This matches the configuration handling in `SWEEnv.from_config()`.

The regression test builds a replay configuration with a startup command, custom timeout, and environment name, then verifies that `_get_env()` passes all three values to `SWEEnv` together with the replay deployment.

Tests:
- `uv run --no-sync pytest -q tests/test_run_replay.py::test_replay_preserves_environment_config`
- `uv run --no-sync pre-commit run --files sweagent/run/run_replay.py tests/test_run_replay.py`

### DIFF
--- a/sweagent/run/run_replay.py
+++ b/sweagent/run/run_replay.py
@@ -174,7 +174,9 @@ class RunReplay:
         return SWEEnv(
             deployment=self.deployment,
             repo=self.config.env.repo,
-            post_startup_commands=[],
+            post_startup_commands=self.config.env.post_startup_commands,
+            post_startup_command_timeout=self.config.env.post_startup_command_timeout,
+            name=self.config.env.name,
         )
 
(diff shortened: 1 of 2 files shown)


anxkhn:

PR #1462, opened 2026-07-15
fix: honor trajectory directory in all run modes

anxkhn:
#### Reference Issues/PRs

Follow-up to #1054, which made batch runs use the configured trajectory directory.

`SWE_AGENT_TRAJECTORY_DIR` defines the shared `TRAJECTORY_DIR`, but single runs and replays previously rebuilt their default output paths from the current working directory. As a result, configuring the trajectory root affected batch runs while these two modes continued writing under `./trajectories`, splitting outputs across different directories.

This changes the single-run and replay defaults to use `TRAJECTORY_DIR`, matching batch-run behavior. Explicit `output_dir` values and the existing per-user and per-run path names remain unchanged.

The regression tests replace the shared trajectory root and verify that both default output paths are derived from it.

Tests:
- `uv run --no-sync pytest -q tests/test_run_single.py::test_default_output_dir_uses_trajectory_dir tests/test_run_replay.py::test_default_output_dir_uses_trajectory_dir`
- `uv run --no-sync ruff check sweagent/run/run_single.py sweagent/run/run_replay.py tests/test_run_single.py tests/test_run_replay.py`
- `uv run --no-sync ruff format --check sweagent/run/run_single.py sweagent/run/run_replay.py tests/test_run_single.py tests/test_run_replay.py`

### DIFF
--- a/sweagent/run/run_replay.py
+++ b/sweagent/run/run_replay.py
@@ -34,6 +34,7 @@ from swerex.deployment.abstract import AbstractDeployment
 from swerex.deployment.config import DeploymentConfig, get_deployment
 from typing_extensions import Self
 
+from sweagent import TRAJECTORY_DIR
 from sweagent.agent.agents import DefaultAgent
 from sweagent.agent.models import ReplayModelConfig
 from sweagent.environment.swe_env import SWEEnv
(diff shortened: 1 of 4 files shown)


anxkhn:

PR #1487, opened 2026-08-02
fix: use native multi-sample requests when supported

### LINKED ISSUE #1484: `n>1` sampling fires N full-prompt requests instead of forwarding `n` to the provider

**Describe the bug**

```python
    if n is None:
    outputs = []
    for _ in range(n):
    return outputs
```

FULL COMMAND (setup + run):

```bash
cd SWE-agent
pip install -e .
python repro.py
```

```python
import litellm

seen = []
    return ModelResponse(
        id="x",
    )

```

**Error message/results**

Observed output:

```
```

**System Information**

- OS: macOS / Linux
- Python: 3.11

**Checklist**

zrh805:
## Summary
- send one native `n` request when LiteLLM reports that the provider supports it
- preserve per-sample requests for unsupported or unknown providers
- preserve caller-specified temperature in the fallback path

## Testing
- `uv run --no-sync pytest tests/test_models.py -q`
- `uv tool run ruff check sweagent/agent/models.py tests/test_models.py`
- `uv tool run ruff format --check sweagent/agent/models.py tests/test_models.py`

Fixes #1484


EvolveAegis:
Independently reproduced this before seeing the PR. On `main`, `LiteLLMModel._query(messages, n=k>1)` calls `_single_query(messages)` in a loop without forwarding `n`, so each iteration re-sends the full history as a separate request: `n` never reaches the provider. Our wire-ledger captured exactly 3 full-history HTTP requests for `n=3` (full request capture, not an estimate). This PR routes `n>1` to a single `_single_query(messages, n=n, ...)` when `get_supported_openai_params` reports `"n"` as supported, and falls back to the per-sample loop otherwise, which is the fix shape we would have proposed. Honest caveat on our measurement: we tested with an Anthropic model, which litellm reports as not supporting native `n`, so the 3-request fallback is the correct post-fix behavior for that backend too. The PR's savings apply to providers that genuinely support `n` (OpenAI et al.), where pre-fix code sent `n` separate full-prompt requests instead of one native-n request. The unit tests plus the pre-existing `n`-forwarding inside `_single_query` cover the fix path we could not exercise end-to-end. Nice bonus that the fallback loop now passes `temperature` through.

PR #1488, opened 2026-08-04
fix: use native provider sampling for n>1

### LINKED ISSUE #1484: `n>1` sampling fires N full-prompt requests instead of forwarding `n` to the provider

daleselaji-dev:
## Summary
- use LiteLLM native n sampling when supported
- retain compatibility fallback for providers without native multi-sampling
- add focused regression coverage

## Problem
LiteLLMModel._query with n=k currently sends the full conversation k times and drops n, multiplying prompt-token and request overhead.

## Solution
Forward n to _single_query first. On UnsupportedParamsError, use the existing loop.

## Tests
- pytest tests/test_models.py (not run: pytest is not installed locally)
- git diff --check passed

## Scope and risk
One dispatch path and one focused test. No output parsing or cost accounting changes.

Closes #1484


bigfish-emily:
Thanks for the fix — the single-request happy path is well covered by `test_query_forwards_native_sample_count`. Two behaviors this PR implements are not yet covered by tests, and I have ready-made regressions for both if useful:

1. **Fallback when the provider rejects `n`** (`UnsupportedParamsError` → one request per sample):

```python
def test_n_sampling_falls_back_when_provider_rejects_n():
    """Providers that reject n>1 fall back to one request per sample."""
    model = get_model(
        GenericAPIModelConfig(
            name="gpt-4o",
            api_key=SecretStr("dummy_key"),
            top_p=None,
            per_instance_cost_limit=0,
            total_cost_limit=0,
        ),
        ToolConfig(parse_function=Identity()),
    )
    responses = [
        litellm.exceptions.UnsupportedParamsError("provider does not support n"),
        _make_mock_response(),
        _make_mock_response(),
        _make_mock_response(),
    ]
    with patch("litellm.completion", side_effect=responses) as mock_completion:
        with patch("litellm.utils.token_counter", return_value=10):
            outputs = model._query([{"role": "user", "content": "test"}], n=3)
    assert len(outputs) == 3
    assert mock_completion.call_count == 4
    # Fallback requests are sent without n.
    assert all(call.kwargs["n"] is None for call in mock_completion.call_args_list[1:])
```

2. **Top-up when a provider silently ignores `n` and returns fewer choices** (the `min(n_choices, len(choices))` path):

```python
def test_n_sampling_tops_up_when_provider_returns_fewer_choices():
    """Providers that silently ignore n get the missing samples requested separately."""
    model = get_model(
        GenericAPIModelConfig(
            name="gpt-4o",
            api_key=SecretStr("dummy_key"),
            top_p=None,
            per_instance_cost_limit=0,
            total_cost_limit=0,
        ),
        ToolConfig(parse_function=Identity()),
    )
    short_response = _make_mock_response()
    short_response.choices = [short_response.choices[0], short_response.choices[0]]
    responses = [short_response, _make_mock_response()]
    with patch("litellm.completion", side_effect=responses) as mock_completion:
        with patch("litellm.utils.token_counter", return_value=10):
            outputs = model._query([{"role": "user", "content": "test"}], n=3)
    assert len(outputs) == 3
    assert mock_completion.call_count == 2
    assert mock_completion.call_args_list[0].kwargs["n"] == 3
    assert mock_completion.call_args_list[1].kwargs["n"] is None
```

Both need `import litellm` added to `tests/test_models.py`. Happy to open a follow-up PR with these if you'd prefer that instead.

PR #1490, opened 2026-08-05
fix: forward n>1 sampling to provider in a single request

### LINKED ISSUE #1484: `n>1` sampling fires N full-prompt requests instead of forwarding `n` to the provider

bigfish-emily:
#### Reference Issues/PRs

Fixes #1484

`LiteLLMModel._query(messages, n=k)` produced `k` samples by calling
`_single_query(messages)` in a loop, so a single logical sampling step re-sent
the full prompt `k` times and never forwarded `n` to the provider.
`_single_query` already supports `n` (it forwards it to `litellm.completion`
and collects `n` choices from one response), so this change:

- forwards `n` to the provider in a single request when the backend supports
  it, eliminating the `n`-times duplication of prompt tokens;
- falls back to one request per sample when the provider rejects `n`
  (`UnsupportedParamsError`), preserving the previous behavior for backends
  without native `n` support;
- tops up with separate requests when a backend silently ignores `n` and
  returns fewer choices than requested, and iterates only over the choices
  actually returned instead of raising `IndexError`.

Tests added in `tests/test_models.py` cover the single-request path, the
rejection fallback, and the short-choices top-up.


bigfish-emily:
Closing as a duplicate: #1488 and #1487 already address the same n>1 sampling issue and were opened earlier with the same design (forward n, fallback for providers that reject it). Keeping the queue clean. The branch and tests remain available if useful.

